PR #55, opened 2017-12-09
Submission navigation updates

### LINKED ISSUE #47: Transition to next step upon 'save and continue'
Instead of going to the 'review' phase (which presents a summary of what they entered), transition directly to the next submission workflow step.  Show the 'review' only if they manually navigate to a past completed step (e.g. from 'funding', they click on 'overview' in the side navigation bar)

birkland:
# Overview
* Transitions directly to next submission step upon `Save and Continue`, instead of the 'review' card
* Adds a 'back' button where appropriate to overview and funding steps
  * This pattern will be carried forward as 'compliance' and 'metadata', etc go live

# How to test
* Check out this PR locally 
  1.  Check out the latest `master` from pass-ember (e.g. `git checkout master` then `git pull`)
  2.  Create a branch from master to try it out in : `git checkout -b birkland-issue-47 master`
  3.  Pull the PR into it `git pull https://github.com/birkland/pass-ember.git issue-47`
  4.  Do `npm install` and then `ember serve`
* Create a new submission via the grants table
  * After clicking 'next' on the DOI stage, you now see a 'back' button that will allow you to go back to DOI entry
    * Note:  the DOI auto-population of title/journal only happens if the fields are empty.  So if you entered a DOI, clicked next, watched it auto-fill, then went back and changed the DOI, it won't change the title and journal 
  * Click `Save and Continue`.  You should now go directly to the funding step
  * If you click `back` at this stage, you'll see the submission overview.  If you click 'next', back to funding
* Create a submission via the submissions table
  * This is pretty much the same as the above.  Notice the `Save and Continue" transitions directly to funding.

Interested parties @htpvu

Resolves #47

### DIFF
--- a/app/templates/grants/show.hbs
+++ b/app/templates/grants/show.hbs
@@ -66,7 +66,7 @@
                 (action "attachWorkflow" newSubmissionObject target=group)
                 (action "saveAndLinkGrant")
                 (action (mut isShowingForm) false)
-                (route-action "transitionTo" "submission.show.index" newSubmissionObject)
+                (route-action "transitionTo" "submission.show.funding" newSubmissionObject)
             ))}}
           {{submission-overview-details submission=newSubmissionObject}}
         {{/workflow-card}}
--- a/app/templates/grants/show.hbs
+++ b/app/templates/grants/show.hbs
@@ -52,6 +52,7 @@
         {{/workflow-card}}
 
         {{#workflow-card step="details" group=group
+            back=(action "back" target=group)
             cancel=(action (queue
                 (action (invoke "rollbackAttributes" model))
                 (action (mut isShowingForm) false)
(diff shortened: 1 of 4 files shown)


htpvu:
Once I progress through the work flow past the Funder step, I'm not able to edit any information. when going back. This is an issue for when I have not enter the funder step previously and want to go back to add it in. 

I wonder if we should get all the steps flushed out, then work through the forward and backward navigation in a more holistic way?

birkland:
Yeah, it's subtle, probably a state diagram is best: based upon where they are; what transitions are available, and what do they do?  Are there times when an 'edit' button is appropriate vs 'back'?

Thankfully, the way the templates are set up it's fairly easy to do whatever we want.  It's just _defining_ what we want.

As far as editing funders are concerned, right now one can only _add_ funders through the UI (except for cancel for unsaved changes).  The consequence of this is that adding a funder might bring in a new policy.  That policy (a type B NIH journal) might require new input on the 'compliance' page and the 'metadata' page.

In any case, if this PR represents an improvement in usability at all (and doesn't introduce any transitions we regret), I'd recommend merging it, and creating new issues to capture additional navigation changes that are necessary.

PR #57, opened 2017-12-09
Laid out  UI elements doe common metadata page

### LINKED ISSUE #56: Layout Common Metadata form for the submission details

htpvu:
# Overview
* Added UI elements to the Common Metadata page according to wireframe
* Included journal's Volume and Issue field

Interested parties @birkland

Resolves #56

### DIFF
--- a/app/templates/components/submission-metadata-common.hbs
+++ b/app/templates/components/submission-metadata-common.hbs
@@ -1,3 +1,4 @@
+<hr>
 <h3>Common</h3>
 <p>
     Please provide additional information about your article/manuscript below:
@@ -15,7 +16,7 @@
           <label for="issue" class="col-sm-2 col-form-label">Issue</label>
 
           <div class="col-sm-10">
-              <textarea id="issue" rows="4" class="form-control" />
+              <input type="text" id="issue" rows="4" class="form-control" />
           </div>
       </div>
         <div class="form-group row">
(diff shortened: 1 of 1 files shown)


app/templates/components/submission-metadata-common.hbs
```diff
@@ -1,9 +1,23 @@
...
+
+            <div class="col-sm-10">
+                <input type="text" id="volume" class="form-control" />
+            </div>
+        </div>
+      <div class="form-group row">
+          <label for="issue" class="col-sm-2 col-form-label">Issue</label>
+
+          <div class="col-sm-10">
+              <textarea id="issue" rows="4" class="form-control" />
```
birkland: This should probably be an `<input type="text"`, as 'issue' is probably only two or three digits long

birkland:
@htpvu This isn't in scope for this ticket, but I was wondering... as far as authors are concerned:  
1. Should they be modelled as entities in the repository (e.g. a person)
2. Should the UI behave in both the "search for an existing person" and/or "enter in information for a new person who doesn't exist in the system" paradigms?  Or are we always _adding_ author information through this form?

htpvu:
about the author entry, for the time being, I think these behaviors are desirable:
- populate those fields based on DOI's metadata, if they exist.
- if DOI doesn't exist, let user enter in the value. 

We only care about corresponding author (because PI may want to contact them about compliance issue). We dont care about other authors, there might be too many of them. I'm sure how that'd work, but i think it'd be good to have corresponding authors modelled as entities.

PR #78, opened 2017-12-11
Issue 62 - Layout issue on the submission's journal selection modal pop up 

jabrah:
Submission workflow buttons re-organized. New buttons are placed at the bottom, as before, but the button toolbar no longer takes up the full width of the component. Any buttons that would overflow the component now break onto a new line. 

These changes apply to buttons:
* View Submissions > Create new Submission workflow
* View Submissions > click an award number > New Submission modal that pops up
* Submission details > Overview, Awards, Metadata, Attachments sections

Examples of new button behavior shown below:

pass-btn-lg

pass-btn-sm

image

### DIFF
--- a/app/templates/components/workflow-card.hbs
+++ b/app/templates/components/workflow-card.hbs
@@ -6,28 +6,22 @@
         </div>
     </div>
 
-    <div class="row">
-        <div class="col-3">
-            {{#if back}}
-            <button class="btn btn-primary btn-small" onclick={{action back}}>Back</button>
-            {{/if}}
-        </div>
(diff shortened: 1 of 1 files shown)


birkland:
This is what it looks like in MS Edge:

It looks like it like to go as narrow as allowable.  

doi

htpvu:
@birkland I missed your comment. I tested the fix out on my machine and did not see this behavior. in chrome. I can reopen the ticket.

birkland:
I'll just create a new issue specifically for MS edge at some point.  We can prioritize it appropriately.

PR #84, opened 2017-12-11
Implement Send button for submissions (issue 54)

### LINKED ISSUE #54: Send an empty POST to the submission service
Right now, there is a button in the UI that kicks off a `doSend` action.  All it does right now is updates a variable; which results in a UI change, via [this code in the submission.js controller](https://github.com/DataConservancy/pass-ember/blob/23361c2a55a48a4fd8d78ac142e7b313769727aa/app/controllers/submission/show/submission.js#L7) and [the corresponding template](https://github.com/DataConservancy/pass-ember/blob/23361c2a55a48a4fd8d78ac142e7b313769727aa/app/templates/submission/show/submission.hbs#L1)

So the idea is to augment that method to send an empty POST to &lt;insert name of the submission python thingy here&gt;, have it do whatever it does, and provide some sort of visual feedback that it worked or not.  

It may just be expedient to hard-code a URL.  Ask @emetsger

Trumbore:
When the Send button of the Submission details page is clicked, a POST
is now made to the URL and port specified in submission.js.  The code
recognizes the OK/fail code that is returned by the CGI server from the
command it has executed (for PASS, this is an FTP operation).

Note that if the initial version of the CGI server is used, the command
will be executed but the return value will not be recognized.  This is
because that version of the server does not allow responses to be sent
for POST requests.  A pull request will be made for an updated version
of the CGI script that is allowed to reply to all requestors, among
other changes.  This may not be the most desirable behavior in the long
term, nor the only possible solution to this problem, but it is needed
at this time.

Resolves #54

### DIFF
--- a/app/controllers/submission/show/submission.js
+++ b/app/controllers/submission/show/submission.js
@@ -1,10 +1,28 @@
 import Controller from '@ember/controller';
 
+// Set these values to specify the server to which the submission POST is made.
+var URL  = 'http://128.84.9.164'
+var PORT = 7000
+
 export default Controller.extend({
     actions: {
-
-        /** Send the submission! */
+        
+        // POST to the server to initiate the FTP operation
         doSend() {
+            var fullURL = encodeURI(URL + ":" + PORT)
+            
+            fetch(fullURL, {
+                method: 'POST'
+            }).then(function(response) {
+                if (response.ok) {
+                    // Command succeeded
+                    console.log("command passed")
+                } else {
+                    // Command failed
+                    console.log("command failed")
+                }
+            });
+                  
             this.set('isSent', true);
         }
     }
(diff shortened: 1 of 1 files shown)


htpvu:
@Trumbore what are all required for me to test this?

Trumbore:
Hanh,

You would need to get the pass-http-cgi repo and change the 
cgi_server.py file so you can run it on your computer.  I set the 
COMMAND to "uptime" and ARGS to "-p".  This will print out the current 
time if the server is called when the button is pressed. Take a look at 
the README.txt file in this project for other setup info, then run the 
script so it is listening for POSTs.

Then, in the source file submissions.js in this PR, set the URL and PORT 
to match the values you are using for the CGI server, above. Then create 
a new submission and after you click Send you should see the time 
printed in the window where the server is running.

     Ben

htpvu:
I have incorporated the changes in this PR with a minor change in URL and Port number in PR #94.

Issue #44, opened 2017-12-07
Logo sometimes does not load


jabrah:
Example: http://localhost:4200/grants/1
Directly entering this URL into the browser, the app will try to load the logo from `localhost:4200/grants/logo.png`. 

Logo image appears as `<img src='logo.png' ... >` in the browser, though in template (application.hbs), it is written as `src='{{rootURL}}logo.png'`. _rootURL_ issue?

markpatton:
But works from http://localhost:4200/grants/. Perhaps some issue with dynamic segments?

PR #114, opened 2018-01-31
Make rootURL available to all pages

karenhanson:
Closes issue #44 
Added rootURL environment variable as extension to default Controller so that it could be used to form the logo path on any page. Previously the logo was only working on paths with one subfolder because `src="logo.png"` was sufficient to find the image in the root directory. Navigating to sub-folders as in `/grants/1` was causing the relative path to fail (../logo.png would have been required).

### DIFF
--- a/app/controllers/application.js
+++ b/app/controllers/application.js
@@ -1,6 +1,8 @@
 import Controller from '@ember/controller';
+import config from '../config/environment'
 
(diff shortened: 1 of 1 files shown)


karenhanson:
I tested out using a helper and that will work for this as well if a lighter-weight solution is preferred

PR #117, opened 2018-02-01
Styled to improve spacing

karenhanson:
- reduced table cell padding
- formatted inputs to improve height and spacing
- added code to prevent x-out button wrapping to next line on page resize
- set minimum height for text areas so they don't display as a single row
- made x-out buttons smaller

### DIFF
--- a/app/styles/app.scss
+++ b/app/styles/app.scss
@@ -46,6 +46,9 @@ a {
 }
 
 /* Buttons */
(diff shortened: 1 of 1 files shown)


karenhanson:
The spacing problems I was seeing are not being experienced by non-Windows users. Closing for now until we can view across browsers/environments

PR #122, opened 2018-02-20
Change look of 'contact author/pi' links in grants/submissions tables

jabrah:
PASS-42

"Contact author/PI" components in the Grants and Submissions tables 
* Now look like normal hyperlinks
* Added on-hover tooltip


htpvu:
@jabrah I loaded this up locally and didn't see any hove-over tooltips.

jabrah:
Which browser were you testing on that wasn't displaying a tooltip?

htpvu:
Chrome

htpvu:
Could you change the tooltip text to this:

- for PI/CO-PIs link: "Click to contact PI"
- for Corr. Author link: "Click to contact corresponding author"

jabrah:
Unfortunately Chrome might still fail to render the tooltip sometimes. At this point, seems the only way around that would be to use a separate tooltip utility (instead of relying on browser behavior)

PR #127, opened 2018-02-28
Adding a page for Submission overview

jabrah:
## PASS-50

This PR adds a new overview page for a submission that displays much of the submission information on a single read-only screen. Roughly follows desired wireframe. Currently mostly functional outside of a few questions below.

### Changes
* New route added to the Ember app at `/submissions/:submission_id` (`/submissions` route still goes to the table displaying submissions).
* First column of Submissions table (header: Article) displays as hyperlink that will navigate to new overview page for the submission.
* Article column in Grants Details page links to this submission overview page.
* New component added for use in `ember-models-table`, simply lets a cell of the table navigate to the submission overview page

### Questions
For @htpvu and maybe @birkland for questions about the Ember model.
* The wireframe calls for a section to display information about attachments, but it doesn't seem like there is any such information in the ember model at the moment. Are attachments handled yet?
* Under _Submission(s)_ section, there can be a list of "preparers". This should show all people that contribute to the submission. Do we track this information?
* In the first box, where does _Publication Date_ come from? Current ember model has _creationDate, updatedDate,_ and _submittedDate_.
* Do we want to show the _last update date_ for each deposit under _Submission(s)_ section?
* The URL for this overview might make more semantic sense under `/submission/:submission_id` instead of `/submissions/:submission_id`. Current behavior is to open the submission editing workflow. Anyone have preferences? Moving this route to a different URL can always be done in a separate PR.

Screenshot


htpvu:
Let's chat on Monday. I think we need to tweak the lay out.

birkland:
Also, I think the word `submission` is being overused.  It would be good to tighten up the terminology.  Internally in the code, we've been using the word `deposit` to describe what happens to individual repositories (e.g. "deposited to PMC").  Whatever we end up deciding, it ought to be consistent between the UI, model, and code for sake of clarity.

htpvu:
@birkland good point.

PR #129, opened 2018-03-08
Add JHU Comliance policy to submission workflow

jabrah:
PASS-65

By default, the JHU Open Access Policy now displays for all submissions. This new component has no logic to decide whether or not to add a new deposit to JScholarship.

* Add _policy-compliance-jhu_ component
* Force JHU policy to be displayed for all submissions

The new test added only checks to see if the component will render successfully.

@birkland might be interested

### DIFF
--- a/app/components/policy-compliance-jhu.js
+++ b/app/components/policy-compliance-jhu.js
@@ -1,0 +1,4 @@
+import Component from '@ember/component';
+
+export default Component.extend({
+});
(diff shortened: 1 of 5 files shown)


htpvu:
John this looks good. Just a few requests for minor updates the the text that I noticed while i was reading through the UI:

- The NIH policy needs a period at the very end of its text. 
- same goes for the JHU policy
- could you add (JHU) to the blue text in the JHU policy card?
- update the JHU policy text so it's easier to read. " Expects ... be accessible in an open access repository. This can be achieved through deposits into existing public access repositories (such as PubMed Central, arXiv, etc.) and/or into Johns Hopkins institutional repository, JScholarship."

PR #130, opened 2018-03-09
Modify submission workflow "repositories" step

jabrah:
PASS-66

## Changes
* Header and explanation text changes
* List of displayed repositories no longer allows users to arbitrarily add other repositories
* List of displayed repos no longer displays JScholarship
* JScholarship now shows under a 'Local' repositories section with a checkbox
  - Checkbox is checked by default
* New deposit is created to JScholarship when the user clicks _Save and Continue_ only when necessary

Screenshot

@htpvu

### DIFF
--- a/app/controllers/submission/show/repositories.js
+++ b/app/controllers/submission/show/repositories.js
@@ -1,4 +1,5 @@
 import Controller from '@ember/controller';
+import { computed } from '@ember/object';
 import { inject as service } from '@ember/service';
 
 export default Controller.extend({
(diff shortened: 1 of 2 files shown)


htpvu:
@jabrah, why does the labels on the required repo say "submitted"?

jabrah:
Most likely because of how the test data is setup. **Submission 1** is setup with two deposits in the test data, both of which have their status set to _submitted_ (example: https://github.com/OA-PASS/pass-ember/blob/master/app/routes/application.js#L297-L301). This behavior is consistent with the `repositories` workflow step in `master`

htpvu:
I dont think it makes sense to display submission status in this workflow. I mean this is when user are just creating the submission. Could you not display the status label here? unless you can see some use for it that i'm missing @jabrah

htpvu:
screen shot 2018-03-10 at 9 24 48 pm
@jabrah I see issue with the required repository not showing. PMC   and PAR should have shown up here,but they didnt.

jabrah:
I definitely saw the same behavior before, but when I tested this again both on this PR branch and on current master branch and was unable to reproduce it.

Issue #74, opened 2017-12-11
Don't keep state in controllers


birkland:
Right now, controllers are used to hold a list of "newly created objects that are supposed to be linked to the submission upon save, or rolled back upon cancel".  This was fast to write, but dangerous as controllers are singletons.  

Maybe use a component instead.  That way cleanup can be tied to the component lifecycle in case the user "navigated away" without saving or cancelling.

Or maybe don't keep an explicit list of new objects at all and somehow use ember-data to reason about links to objects that haven't been committed.

htpvu:
This ticket may be out of date after the UI code update.

PR #131, opened 2018-03-12
Submission workflow Repositories list population issue

jabrah:
## Issue description:
* Sometimes necessary repositories are not populated in the list. Say a submission has one NIH award that will require a deposit to PubMed. In this case, PubMed may not appear in the Repositories list.
* If a submission already has deposits to any repository, they _should_ appear in the Repositories list anyway. If no deposit has yet been created for this award in the submission, it will not appear in the Repo list in its current state.
* If a user goes directly to the Repositories step for any submission that does not have deposits created in the DB, nothing will appear. In this case of direct navigation, only previously created deposits are shown. A user must go through the workflow from at least Policies step, then click Save and Continue in order for PASS to create new deposits.
* A submission is created with an award. User navigates away at the Policies step, then uses the browser Back button and clicks the Save and Continue in the Policies step. Now PubMed will appear twice in the Repositories list, where it should only appear once.
* A submission is created with two awards that follow the same compliance policy. In the Polices tab, the same policy is now displayed twice, where it should only appear once.

## Code changes:
* Fix issue in the NIH policy component logic and how it dealt with the Ember component lifecycle
* Prevent duplicate polices from appearing and from generating duplicate deposits

## Notes on fix:
* User still cannot navigate _directly_ to the submission workflow **Repositories** step. For the correct repositories to appear, it seems that the user must progress through the submissions workflow. More specifically, a user MUST use the "Save and Continue" button from the Policies step in order to reach the Repositories step.
  * The "Save" function in the Policies step (which is triggered from the "Save and Continue" button) is responsible for triggering the registration of each policy. These policies are what create new deposits, so this step is necessary in order to correctly populate the Repositories step.
* The change that keep duplicate polices from appearing and duplicate deposits from being created is a simple workaround and possibly stems from maintaining state in the controller. See [Issue 74](https://github.com/OA-PASS/pass-ember/issues/74). This needs to be addressed in a later PR.

Testing must be done manually in the Submission workflow, both with currently existing and new submissions.

### DIFF
--- a/app/components/policy-compliance-nih.js
+++ b/app/components/policy-compliance-nih.js
@@ -44,14 +44,10 @@ export default Component.extend({
         var self = this;
         register(function () {
             // A never needs deposit, C, D does.  B depends on user input.
-
-            if (pmcMethod === 'A') {
-                self.set('needsDeposit', false);
-            } else if (!pmcMethod || pmcMethod === 'C' || pmcMethod === 'D') {
-                self.set('needsDeposit', true);
-            }
-
-            if (self.get('needsDeposit')) {
+            let mustDeposit = !pmcMethod || pmcMethod === 'C' || pmcMethod === 'D';
+            let answerFromUser = pmcMethod === 'B' && self.get('needsDeposit');
+            
+            if (mustDeposit || answerFromUser) {
                 return self.get('store').createRecord('deposit', {
                     repo: 'PMC',
                     status: 'new'
--- a/app/controllers/submission/show/compliance.js
+++ b/app/controllers/submission/show/compliance.js
@@ -18,7 +18,9 @@ export default Controller.extend({
 
             while (depositGenerators.length) {
                 let deposit = (depositGenerators.pop())();
-                if (deposit) {
+                // Don't add the deposit if it goes to a repository that is already present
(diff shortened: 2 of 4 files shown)


htpvu:
It does work better. There is still a bug, that if you can't fix before Wednesday, I can merge this and let sayeed know of the bug to avoid demoing it: If you add some award, go to the repositories page --> see repos. then go back, remove all rewards, then go back to the repositories page --> see that one or more of those required repos are still there. If you go back and forth a couple time without adding the awards back in, the required repos will go away, but not when you get there the first time after removing awards. Ping me if the reproducing steps are clear.

jabrah:
So it seems like this bug is a direct result of maintaining state information inside the controller ([issue 74](https://github.com/OA-PASS/pass-ember/issues/74)). The compliance controller keeps a running list of deposit generators that is supposed to include functions that create deposits for the current submission (https://github.com/OA-PASS/pass-ember/blob/master/app/controllers/submission/show/compliance.js#L5).

When navigating back and forth in the workflow, this list is not properly updated as one would expect. The recent commit provides a crappy workaround for this. This needs refactoring to actually address the issue.

PR #164, opened 2018-04-26
Implement NIH checking

cameronblandford:
If the user is using the NIH policy and is using a method A journal, they will be alerted as such with the proper text. If the user is not using a method A journal, they will be asked if they have paid the respective submission fee to their journal/publisher. If so, Pubmed Central is removed as a repository. If not, the user continues as normal.

Note:

* This does not yet hide the Pubmed Central metadata form because the metadata form is taken from the policy, not the repository. This will be added soon.
* Picking a radio button option and clicking next twice through the next two steps, then going back and changing your radio button input to the other option causes undefined behavior. This will be addressed soon.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,6 @@
 language: node_js
 node_js:
 - node
-- 7
-- 6
 sudo: required
 addons:
   chrome: stable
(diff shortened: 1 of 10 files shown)


Rytiggy:
LGRTM 🎉

PR #196, opened 2018-04-27
Fix/file persistence between submissions

cameronblandford:
* Reset files array each time component is loaded
* Allows repositories to not have formSchemas - it simply skips them.
* Updates wording on ERIC form. (Fixes https://github.com/OA-PASS/pass-ember/issues/197)


Rytiggy:
LGRTM

PR #198, opened 2018-04-27
Enable file validation

cameronblandford:
`Next` button is disabled if no files are present.
Fixes https://github.com/OA-PASS/pass-ember/issues/191.


Rytiggy:
LGRTM

PR #201, opened 2018-04-30
Branding update

Rytiggy:
- Increase fontsize to 1rem for body text elements
- Adjust text colors for better contrast
- Better postion logo


cameronblandford:
Looks good to go!

PR #203, opened 2018-05-01
Update docker to include search support

markpatton:
Adds Elasticsearch and pass-indexer containers to docker. Objects added to Fedora will be automatically indexed. See https://github.com/OA-PASS/pass-indexer for more info.

Update to latest ember-fedora-adapter with support for search. 

To test, docker-compose up, wait for the indexer to find both Fedora and elastic search, and then run the app. There will be some errors due to the Ember model not matching the current pass context when test data is created.

### DIFF
--- a/.elasticsearch.yml
+++ b/.elasticsearch.yml
@@ -1,0 +1,7 @@
+http.host: 0.0.0.0
+
+http.cors.allow-origin: http://localhost:4200
+http.cors.enabled: true
+http.cors.allow-credentials: true
+http.cors.allow-methods : OPTIONS, HEAD, GET, POST, PUT, DELETE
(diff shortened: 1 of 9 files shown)


cameronblandford:
Once you get those two tests passing, we're good to go!

markpatton:
@cameronblandford The errors are due to the acceptance test. That loads the application route which creates the test data. @jabrah is working on fixing up the test data, but there is still a question about dealing with acceptance tests. The user will have to docker-compose up before running them. Travis will have to be updated to use docker. I can do the travis update, but should I?

markpatton:
@cameronblandford I added in Travis support. Tests will still fail until Johns work.

PR #204, opened 2018-05-01
Update Ember internal model

jabrah:
Various Ember model updates to bring it in line with `pass-data-model` and make the application compatible with Fedora.

Note: Mirage is getting in the way of `user` objects, so they are not able to be added to Fedora. Until this is fixed, things like _Corresponding Author_ column in the Submissions tab will be broken.


markpatton:
The test on Travis dies complaining about set which is new to the latest fedora adapter. Should check that dependency.

markpatton:
Actually the problem is that mirage is configured for User which has a set attribute only the fedora adapter knows about. One possible solution is to temporarily add a set transform.

markpatton:
Oddly all of this passes fine locally.

PR #205, opened 2018-05-02
update landing page text

Rytiggy:
landing page

## issues 
- Agencies do not auto rotate, after they are clicked they start rotating.


jabrah:
I think one of the changes my PR includes will fix the Travis problem here: I have one extra parameter in `.env`

jabrah:
The new landing page looks nice!

Issue #200, opened 2018-04-30
Consider changing up the font for readability


htpvu:
Feedback from sayeed (and I) is the current font used is hard to read (too small, not enough contrast). Can we change this up in ways that still conform to the style guide, but is easier to read.

Rytiggy:
## Todo to improve readability
- Increase fontsize to 16px for body text elements 
- Adjust text colors for better contrast
- Change title font to Arnhem Pro
- Better postion logo

Rytiggy:
https://github.com/OA-PASS/pass-ember/pull/201

htpvu:
@Rytiggy is the blue used for hyperlink the darkest blue yet? it's still pretty hard on the eyes for lack of contrast with white background.

Rytiggy:
@htpvu This should be more readable now. 
https://github.com/OA-PASS/pass-ember/pull/224

![image](https://user-images.githubusercontent.com/17414914/39596601-e1bd3e86-4ee0-11e8-8b8f-fc63b34d2045.png)

htpvu:
looks good.

PR #207, opened 2018-05-02
update to meet JHU branding guidelines

Rytiggy:
image
https://github.com/OA-PASS/pass-ember/issues/200


jabrah:
Looks good to me

PR #214, opened 2018-05-03
update text and images for landing page

Rytiggy:
localhost_4200_welcome 1


app/templates/index.hbs
```diff
@@ -13,7 +13,8 @@
 
     </div>
   </div>
-  <h1  class="font-weight-light mt-4 text-center m-4">Agencies</h1>
+  <h1  class="font-weight-light mt-4 text-center m-4">PASS supports these agencies' and institutions'<br> public and open access policies</h1>
```
cameronblandford: Maybe here have `Supported Agencies and Institutions` as the header, and then have a h2/h3 or p tag saying the rest?

cameronblandford: Also, unrelated, but maybe add a "?" to the Interested h1 element?

htpvu: i agree with cam's suggestion

cameronblandford: Looks good pending tests!

PR #216, opened 2018-05-03
Fix a series of bugs related to model update

cameronblandford:
* Sometimes repositories don't load.
* Policies don't load.
* Journals aren't created properly.


Rytiggy:
LGRTM 🎉

PR #224, opened 2018-05-03
change link color to darker blue

Rytiggy:
https://github.com/OA-PASS/pass-ember/issues/200


cameronblandford:
LGRTM pending tests!

Issue #237, opened 2018-05-04
Grants table # column doesn't populate on fresh reload


jabrah:
If you load app at `/grants` to see the table of grants, the `#` column shows a value of 0 for all grants, which is not correct. Only when you load submissions will they be counted here. This has to do with loading relationships for this route.

cameronblandford:
The solution for this for now can just be loading the submissions inside the grant route.

PR #243, opened 2018-05-04
Load submissions in grants index every time

cameronblandford:
Fixes https://github.com/OA-PASS/pass-ember/issues/237


Rytiggy:
LGRTM

PR #246, opened 2018-05-04
Get submissions POSTing

cameronblandford:
## Included:

* Switch from file POJO to file ember record
* Get submissions POSTing correctly
* Add in file role dropdown, with first file defaulting to `manuscript` and the rest defaulting to `supplement`.
* Fully regression tested

### DIFF
--- a/app/templates/components/workflow-files.hbs
+++ b/app/templates/components/workflow-files.hbs
@@ -1,68 +1,62 @@
 {{yield}}
 <style>
-.table-text {
-  display: inline-block;
-  overflow: hidden;
-  white-space: pre-wrap;
-  width:200px;
-}
+  .table-text {
+    display: inline-block;
+    overflow: hidden;
+    white-space: pre-wrap;
+    width: 200px;
+  }
 </style>
 <p>
-   Attaching manuscript/article files to this submission.
+  Attach manuscript/article files to this submission.
 </p>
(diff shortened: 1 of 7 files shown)


app/templates/components/workflow-files.hbs
```diff
@@ -1,62 +1,62 @@
...
+          <tr>
+            <td>
+              <p class="table-text">{{file.name}}</p>
+            </td>
+            <td class="text-center">
+              <select onchange= {{action (mut file.fileRole) value="target.value"}}>
+                <option value="supplement" selected={{eq file.fileRole "supplement"}}>Supplement</option>
+                <option value="manuscript" selected={{eq file.fileRole "manuscript"}}>Manuscript</option>
+                <option value="table" selected={{eq file.fileRole "table"}}>Table</option>
+                <option value="figure" selected={{eq file.fileRole "figure"}}>Figure</option>
```
Rytiggy: Maybe we should add an other option here as well so we cover everything?

Rytiggy:
LGRTM

app/templates/components/workflow-files.hbs
cameronblandford: @htpvu wanted it to be ENUM-style to purposefully limit the options, I think. Plus, supplement is basically "other".

PR #248, opened 2018-05-04
Minor Fixes

cameronblandford:
* Standardize capitalization of grant statuses
* Remove notifications from navbar


Rytiggy:
RTM pending tests

Issue #254, opened 2018-05-05
User should be prevented from jumping ahead of the workflow if required data is not yet provided in previous step by click on the progress bar. 


htpvu:
<img width="964" alt="screen shot 2018-05-05 at 2 48 06 pm" src="https://user-images.githubusercontent.com/11050142/39666550-715c5a02-5073-11e8-8d73-f137a4e8a5cd.png">

PR #261, opened 2018-05-07
Add progress-checking to submission nav

cameronblandford:
Prevents people from skipping ahead farther than they've clicked `next` buttons for. Adds `disabled` attribute to links in submission nav that have not been reached yet in the workflow. Fixes https://github.com/OA-PASS/pass-ember/issues/254.


Rytiggy:
LGRTM

PR #305, opened 2018-05-11
Remove mocked user service from Mirage

jabrah:
This change unblocks `user` objects in Fedora, so make sure to refresh your Docker environment. New implementation is still mocked, but it simply picks a user from Fedora (picks out `hvu` now)

**Has not been regression tested** unfortunately due to personal time constraints, but we can do code review at least


cameronblandford:
This looks good to me, pending regression testing.

cameronblandford:
@jabrah can you add a test somewhere to increase coverage so it'll pass?

PR #309, opened 2018-05-14
Fix #308

### LINKED ISSUE #308: Update handling of DOI handling
DOI look up is failing in a secured environment. - need to be able to make request using https protocol
Look up of DOIs need to support multiple DOI encoding 
http://www.doi.org/syntax_table.html
naked doi: 10.1021/acsmedchemlett.7b00397
https doi:  https://doi.org/10.1109/5.771073
http doi:  http://doi.org/10.1109/5.771073

cameronblandford:
* Definitively fixes the url-encoded DOI lookup issue with https://github.com/OA-PASS/pass-ember/issues/308
* Should fix the the security issue with DOI lookup in https://github.com/OA-PASS/pass-ember/issues/308
* Adds a basic test for the journal model with helpful example logic


Rytiggy:
yay lgrtm

PR #312, opened 2018-05-14
Add sorting and filtering back into ember-models-table for submissions

cameronblandford:
#Notes

Award, repository, and title columns are now searchable and sortable. This was done by adding computed fields to submissions - arrays of strings - containing the text for each of those categories, and setting the relevant column's `propertyName` to that property.

## To Do

- [x] Confirmed that OAP repocopy sorting + filtering isn't a hard requirement, as it's prevented by the removal of the `deposits` attribute from the `submission` model.
- [x] Regression-tested

### DIFF
--- a/app/components/submission-status-cell.js
+++ b/app/components/submission-status-cell.js
@@ -1,0 +1,15 @@
+import Component from '@ember/component';
+
+export default Component.extend({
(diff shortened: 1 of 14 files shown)


Rytiggy:
LGRTM pending tests

PR #315, opened 2018-05-14
Add support for http(s) URLs and dx.doi URLs in validation and lookup

cameronblandford:
As per @htpvu's suggestion


Rytiggy:
Very nice! RTM

Issue #322, opened 2018-05-15
Update grant selection feature in submission workflow


htpvu:
In Grant step of the workflow, populate the drop down with logged-in user's grants (on which he/she is the pi) - sort by award number.

jabrah:
This will likely be rolled into a future update to Autocomplete (#279), as the updated `pass-indexer` should include the ability for the search service itself to do the filtering.

htpvu:
I don't see this restriction happening. I'm still seeing all the grants - where I should only see 5.

On rev. dd4d6985

In fact, can we please make the drop down to be populated with user grant. so he/she can select from the drop down instead of having to search? that would will probably remove the need to use auto complete?

jabrah:
whoops! sorry if I mislabeled this ticket. My intent was that this ticket needs verification in the unmerged PR (https://github.com/OA-PASS/pass-ember/pull/292) - so code is not yet in `develop`. Same label situation in #279

htpvu:
this is done in a different ticket

Issue #280, opened 2018-05-09
Search integration needs an efficiency pass


jabrah:
Current implementation of `Store.query` in the UI is very inefficient, resulting in many search requests for some pages. In some of these cases, the multiple search requests can be consolidated into a single search request that is then parsed by the UI.

Non exhaustive list of places to investigate:
* `grants.index`
* `submissions.index`
* `grant.detail` when populating the associated submissions table

cameronblandford:
@jabrah what's the status of this ticket?

jabrah:
Things could be optimized further, but I think it is in an OK state

PR #337, opened 2018-05-16
Show only content relevant to logged in user in the Grants and Submissions tables

jabrah:
Work in progress PR to update various views to show only content relevant to the currently logged in user.

- [x] Grants table should show only grants where currently user is the PI #142 
- [x] Submissions table should show only those submissions where the current user is the managing user (`submission.user === current_user`)
- ~~Grant-picker in the submission workflow should contain only grants where the current user is the PI #322~~ Will be rolled into separate PR

Will include an efficiency pass for using the search service (#280)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ before_script:
   - docker-compose up -d
   - bash .wait.sh
 script:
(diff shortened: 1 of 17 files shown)


cameronblandford:
LGRTM 👌👍🙌

PR #347, opened 2018-05-18
Update Grants details page

jabrah:
* Mailto link removed from PI and coPIs
* Submission table updated to be in line with updates to table in Submissions page
* Added test for grant details page


Rytiggy:
This looks good to me! RTM

Issue #206, opened 2018-05-02
Agencies do not auto rotate on landing page, after they are clicked they start rotating.


Rytiggy:

Rytiggy:
This is working on the static stite now that they are static pages. see on the index page
https://rytiggy.github.io/pass-ui-static/index.html

PR #354, opened 2018-05-22
Resize agencies logos 

Rytiggy:
## Fixes https://github.com/OA-PASS/pass-ember/issues/206
- resize logos 
screen shot 2018-05-22 at 2 32 20 pm


cameronblandford:
lookin' good!

PR #364, opened 2018-05-25
Remove most Store.findAll function calls

jabrah:
After this code change, `Store.findAll` is only present in `submission/new` route.


jabrah:
Superseded by another PR

PR #292, opened 2018-05-10
Using Elasticsearch autocomplete service for journal and grants pickers

jabrah:
Needs a good QA pass!

Note this implementation has different behavior compared to the old mocked "autocomplete." This implementation will only offer suggestions where the searched fields are prefixed with the user input. Grants are _not_ yet filtered by logged in user.

Example:
* User types `eng`, attempting to find Advances in Engineering Education. Autocomplete returns no results.
* User types in `adv`, Advances in Engineering Education now appears

**Questions**: 
* Does the Elasticsearch autocomplete service need the user credentials?

PR #369, opened 2018-05-25
Destroy Store.findAll

markpatton:
Builds on John's work to get rid of findAll usage.
Also modify most Store.query to have a larger result size.
Later on we will have to figure out paging.

Two uses of findAll remain in app/components/find-grant.js and app/components/find-journal.js which will be addressed in the autocomplete pr.


jabrah:
**Note:** there are TWO `Store.findAll` calls left in the codebase, however they are eliminated by the autocomplete changes PR https://github.com/OA-PASS/pass-ember/pull/292

Issue #366, opened 2018-05-25
Update text for the "thanks" page


htpvu:
To:

> Thank you for submitting your publication through PASS. Detailed summary of the submission you just made can be seen here. For list of all submissions associated to your account, please see here.

htpvu:
The links are point to the wrong places: For submission details, link user to the submission details page of the submission they just submitted. For list of submissions, link user to the submissions table page

PR #375, opened 2018-05-29
Implement environment-dependent authorization for file-uploads

cameronblandford:
* Adds environment-checking for headers when uploading files
* Updates text on `thanks` page (#366)


Rytiggy:
LGRTM accepting *1 failing check*

PR #398, opened 2018-05-30
Update docker configuration [WIP]

markpatton:
This runs ember server in a container, `ember`, on the code along with the rest of the stack need for shibboleth.

In order to test, do a docker-compose up and wait for containers to settle.
Then visit https://pass/. Note that pass must be aliased to localhost.

TODO:
[ ] Update README
[ ] Integrate real data and user service
[ ] Get travis working


markpatton:
Going to do this differently. No longer needed.

Issue #352, opened 2018-05-22
For every place where the UI blocks user from proceeding in the workflow...


htpvu:
In form user of why they can't proceed. make required action more obvious: eg file selections, required fields, click on ED link

htpvu:
I can see the warning on the files step. I would have like some inline warning, such as highlight the section where required user's input is missing. But i'm willing to go to user testing with the current method of warning. 

However, i can't tell if the warning is working on the submit step. Submit button is way in the bottom, I can't feel if the warning box only appears in the header, which is out of sight, so i can't see it, OR if the warning box didnt' show up at all. 

<img width="1164" alt="screen shot 2018-05-25 at 1 13 18 pm" src="https://user-images.githubusercontent.com/11050142/40557347-cd44bbb6-601d-11e8-826f-23abb0eeab20.png">

htpvu:
as discussed, move popup modal about un-signed deposit agreement to the Jscholarship meadata table,

htpvu:
@Rytiggy i think we need to talk more about the orange box in the review step warning about clicking on ERIC link, themessage is very vague

htpvu:
on rev. 9792cf64, I was not prompt to agree to the deposit agreement. So I didn't and the submission went anyways. But it went with an empty list of Repositories (as seen in the submission details page).

So two things expected that didn't happen here:
- user needs to be prompted to agree to deposit agreement before leaving that page
- If they didn't agree, and the repository list is empty we may want to handle it differently, like as user to exit the workflow. This is related to ticket #426 that @cameronblandford is working on. Let's chat about it tomorrow? 

<img width="1151" alt="screen shot 2018-06-05 at 10 09 53 pm" src="https://user-images.githubusercontent.com/11050142/41011929-379f5f64-690d-11e8-9fd3-c428edf584b7.png">

Rytiggy:
@htpvu you should see this alert if the user does not agree to deposit.

Granted, Ok and cancel should be changed to ok and agree to deposit.

![screen shot 2018-06-06 at 2 18 25 pm](https://user-images.githubusercontent.com/17414914/41057234-96212702-6994-11e8-9616-99b9e0566d80.png)

Rytiggy:
@htpvu the schema for JScholarship needs to be updated, as soon as this is done, it will be fixed. 

```json
 
{
        name: 'JScholarship',
        url: 'https://jscholarship.library.jhu.edu/',
        formSchema: `{
[...]

PR #404, opened 2018-05-31
Better user feedback

Rytiggy:
## https://github.com/OA-PASS/pass-ember/issues/352
- Add warning to indicate to user to submit to external submission portals
- Fix grant table create new button not keeping grant when you create new submission 
- Add validation for agreeing to deposit and external submissions
- Regression tested


cameronblandford:
:+1:

PR #418, opened 2018-05-31
Replace example.org in context URL with oapass.org

karenhanson:
This line may require an update at some point, but not sure what to: https://github.com/OA-PASS/pass-ember/blob/55f43dcd3a400e2548f2cf922d89fcca29c80b2b/app/components/workflow-files.js#L53

### DIFF
--- a/.env
+++ b/.env
@@ -1,8 +1,8 @@
 # Adapter integration test config:
 
 FEDORA_ADAPTER_BASE=http://localhost:8080/fcrepo/rest/
-FEDORA_ADAPTER_CONTEXT=http://example.org/pass/
-FEDORA_ADAPTER_DATA=http://example.org/pass/
+FEDORA_ADAPTER_CONTEXT=http://oapass.org/ns/pass#
+FEDORA_ADAPTER_DATA=http://oapass.org/ns/pass#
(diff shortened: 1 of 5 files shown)


markpatton:
The same changes will also need to be made to .docker/shib/docker-compose.yml

.env
```diff
@@ -16,9 +16,9 @@ FCREPO_STOMP_PORT=61613
 FCREPO_LOG_LEVEL=INFO
 FCREPO_ACTIVEMQ_CONFIGURATION=classpath:/activemq-queue.xml
 
-COMPACTION_URI=http://example.org/pass/
-COMPACTION_PRELOAD_URI_PASS_STATIC=http://example.org/pass/
-COMPACTION_PRELOAD_FILE_PASS_STATIC=/usr/local/tomcat/lib/context-2.0.jsonld
+COMPACTION_URI=http://oapass.org/ns/pass#
```
birkland: This is all in fcrepo image, so is unnecessary unless pass-ember devs are making local changes to the context.

htpvu:
Tested it with pass-docker, works to me. @cameronblandford or @jabrah please have a look at the failed test and see if this PR could be merged without nasty consequences. Thanks!

PR #419, opened 2018-06-01
Add docker configuration with support for Shibboleth

markpatton:
* Add a docker configuration in .docker/shib which runs the local code in the ember container. See the updated README for more info.
* Consolidate config files used by original docker config in .docker

In order to test, in .docker/shib try docker-compose up and then visit https://pass/.
Then shut all that down and prune and also try docker-compose up in the toplevel directory like before.

Note that acceptance tests now fail in the original docker config although it remains untouched.


cameronblandford:
LGRTM

PR #424, opened 2018-06-01
Integrate user service + autocomplete service

jabrah:
First cut at integrating the user service into the Ember app. This will finally allow the full `pass-docker` stack to be used with the UI, and allow us to use real data. The PR to include an autocomplete service is also folded into this change because it is needed to fully remove any `Store.findAll` calls.

The new docker environment has not yet been updated in this change, so Ember changes will not be seen as they are made. Instead, the old `pass-docker` workflow must be followed in order to see changes:

* Make code changes in Ember
* Commit changes, push to Github
* Modify `pass-docker/.env` to point to the appropriate Github repo and commit
* Change the Ember container tag in `pass-docker/docker-compose.yml`
* Rebuild the ember container in `pass-docker` with `docker-compose build ember`
* Bring things up

### Testing

In order to test, you have to follow the steps above. You will need to clone `pass-docker` and check out [`pr/42`](https://github.com/OA-PASS/pass-docker/pull/42) (`birkland:bootstrap-assets`) . `pass` must be aliased to `localhost` on your machine. Similar changes will be made to `pass-docker` when appropriate (i.e. sometime after the PR is tested and merged). Tests will be run in the pass-docker environment using the pr/42 codebase.

First, `docker-compose pull` to make sure you have the latest images.

**`.env`** in `pass-docker` needs to be modified:
```
EMBER_GIT_REPO=https://github.com/jabrah/pass-ember # Should point to code for this PR
EMBER_GIT_BRANCH=d1aba69808aa9531abc349c226e9068acc08f514 # Should point to the tip of the PR

USER_SERVICE_URL=https://pass/pass-user-service/whoami # New property needs to be added
```

**`docker-compose.yml`** in `pass-docker`
I have been changing the ember container tag to something random to avoid problems.
```
ember
  image: oapass/ember:moo # I only use moo because I am not intending on publishing the image while testing :)
```

In `pass-docker`, run the command `docker-compose build ember` to build. Then do `docker-compose up` to bring stuff up. When things settle, in a browser, navigate to `https://pass`. This should bring up a login screen. Credentials `username: faculty1 pass: moo` can be used.

The `pass-docker` PR contains more relevant information.

### Known Issues
* User `staff2` can login with Shib, but is denied access to PASS. The User service returns a `401` code as expected, but this code is not handled elsewhere
* Autocomplete for journal names is not yet working


htpvu:
I tested this. it works as expected. But I'm holding off from merging because it's dependent on https://github.com/OA-PASS/pass-docker/pull/42 merging. pass-docker is currently a bit unstable, so that can't be merged just yet. Will find out more about the state of pass-docker at Monday morning meeting.

PR #465, opened 2018-06-08
Bump shib env image versions to add new test users

jabrah:
Adds new test users with appropriate grants:
* `nih-user`
* `ed-user`
* `usaid-user`

Remember to bring down the environment and pull new images. Only `fcrepo` and `ldap` images are updated

Verified locally


Rytiggy:
LGRTM

PR #472, opened 2018-06-09
update submission detail page if source == other

Rytiggy:
screen shot 2018-06-09 at 3 28 48 pm

### DIFF
--- a/app/templates/submissions/detail.hbs
+++ b/app/templates/submissions/detail.hbs
@@ -54,8 +54,11 @@ table td {
           <tbody>
             <tr>
               <td class="max-width">Status</td>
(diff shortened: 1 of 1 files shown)


htpvu:
Initial request: 
- if a submission has source of “other” - then 
    - display status as “This submission did not originate from PASS.” 
    - display Submitter  as “This submission did not originate from PASS”
    - display Files as “We do not have a copy of manuscript files because this submission did not originate from PASS”
    - hide Deposits

----

@Rytiggy I would like to make this change to the status text please:

if source=other, submitted=false --> status= "You funder is aware of this publication and is expecting the deposit of your manuscript."
if source=other, submitted=true --> status = "A manuscript submission has been made outside of PASS"

PR #484, opened 2018-06-09
Fix 404 error when reloading ember URLs through shib

birkland:
By default, Apache (`proxy` and `sp` images) would automatically return a 404 for any URL that had an encoded slash.  This was problematic when hitting the browser's reload button, for example.   This PR updates the images so that these URLs are passed on unmolested.


cameronblandford:
LGRTM

Issue #466, opened 2018-06-09
restyle the respository section in the review page, and possibly the details page. 


htpvu:
<img width="464" alt="screen shot 2018-06-08 at 8 27 33 pm" src="https://user-images.githubusercontent.com/11050142/41185862-12149a02-6b5b-11e8-8659-c8e580c5ed48.png">

htpvu:
a couple things we should change: 

- the width of the column containing "repositories" should never be so narrow that it needs a second line.
- Put the explanation text in a tooltip: "These are the repositories that PASS will send your submission to"

Rytiggy:
![screen shot 2018-06-09 at 8 46 10 pm](https://user-images.githubusercontent.com/17414914/41197283-4496c45c-6c26-11e8-8132-8cea00124cac.png)

cameronblandford:
LGRTM

htpvu:
verified no rev. 2f8aa8d5

PR #486, opened 2018-06-10
Fix/review page repos 

Rytiggy:
https://github.com/OA-PASS/pass-ember/issues/466


app/components/workflow-basics.js
```diff
@@ -46,7 +46,8 @@ export default Component.extend({
   didRender() {
     this._super(...arguments);
     this.send('validateDOI');
-    this.send('lookupDOI');
```
cameronblandford: What kind of bugs? This is required for having prepopulated DOI stuff when editing existing submissions

Issue #521, opened 2018-06-11
Journal autocomplete on "1. Basics" step does not work as expected


karenhanson:
If I do not search for a DOI and instead manually search for the journal , the auto-fill does not quite work as expected. For example, there are 65 matches for Nature in elasticsearch: 
http://localhost:9200/_search?q=@type:Journal+name:Nature&default_operator=AND&pretty&size=30 
If I attempt to type "Nature" (mixed case) in the Journal field on the "1. Basics" step of the workflow, no matches are found on "N" "Na" "Nat"... etc. all the way to the full word.  If I enter "nature" lower case, it matches but I must enter the full word before they appear.  Similarly, there are many journal names that start "Journal of..." no matches are found on "Journal". If I type lower case "j" some matches come up but these disappear if I type "jo", then reappear if I finish the word to "journal" all lower case. It's not clear to me what is causing this behavior, it looks like case sensitivity plus some other element that I can't figure out.

cameronblandford:
This is an elastic-search issue, not an issue on the UI side.

karenhanson:
Seems like this could be the same as #279 actually but wasn't sure.

birkland:
@karenhanson I can work on updating the `assets` container with fixed journal records

jabrah:
@karenhanson The current Journal picker uses a temporary method of retrieving Journal suggestions that uses normal Elasticsearch queries. Current behavior will match only full words.

This was done because the Journal data needs to be updated as @birkland just mentioned. Once this is done, autocomplete can re-implemented in the UI, which should have the desired behavior. For user testing this week, we just needed something that kind of worked.

karenhanson:
This is fixed in the latest pass-docker

karenhanson:
Verified on rev. 7f2a26bd

karenhanson:
Something still not right here. I can see the journal "Nature" in elasticsearch, but when I type "Nature" it doesn't come up as an option.

birkland:
It looks to me like it displays a list of up to N results, even if the total number of hits is much higher.   Maybe there needs to be a sort order?

karenhanson:
Yes, perhaps some kind of boost for exact match. There are 65 matches for "nature"

karenhanson:
@htpvu I believe this is fixed.

PR #534, opened 2018-06-11
Update assets to include proper journal data

birkland:
This uses the proper `journalName` field for journal names, and should have the `_suggest` fields in the index.

Feel free to incorporate this into a PR for the fix to #521 or #239 and close this PR, or merge this first then fix.  It is unclear if merging this will break things due to the workaround. 

/cc @jabrah


jabrah:
This change included in PR https://github.com/OA-PASS/pass-ember/pull/535

Issue #538, opened 2018-06-12
Change policy subtitle "Expects deposit into <repo name>" to "Requires deposit into <repo name>


htpvu:
in step "Policies" of submission workflow

jabrah:
Should this also effect JHU policy? This one currently says "Expects deposit into an open access repository"

karenhanson:
verified in 7f2a26bd

Issue #537, opened 2018-06-12
Update to the repositories step in submission workflow


htpvu:
1. Change "Local Repositories" in the repositories step to "Institutional Repositories"
2. After repository name , attach the name of the agency whose policy will be satisfy by deposit into this repository: eg: 
- PubMed Central - required by National Institutes of Health
- PubMed Central - required by Center for Disease Control and Prevention
- JScholarship - required by Johns Hopkins University
3. If Jscholarship is the only target repository, hide the checkbox, and dont give user a way to opt out of it. 

Consideration: the desired effect is to convey that Jscholarship deposit is optional if user is submitting to PMC. But we need to balance between a tailored message and keeping the logic of loading policies & repository extensible and dynamic. 

We should chat and brain storm some about (3)

karenhanson:
@jabrah tested this, 1 & 2 work, but the checkbox needs to be hidden for (3) unless there has been an offline discussion to leave it visible but grayed out.

htpvu:
Greying out is ok, if it is too much work for the time being. But we prefer hidden text bow.

karenhanson:
verified in 566f344

PR #542, opened 2018-06-12
Adjusting text in submission workflow

jabrah:
#537 #538 

## Testing
* Login as `nih-user`
* Navigate to new submission workflow to start a new submission
* In Grants step, enter random grants
* In Policies page, **Policy cards should have sub-title that says something along the lines of: _Requires deposit into ..._**
* In Repositories step, **each repository should say _- required by &lt;some funding agency&gt;_**
* Must also verify that the end artifacts look good - so finish out the submission workflow and check the submission info - the resulting submission should have the expected Repositories listed in its details page

### DIFF
--- a/app/components/workflow-repositories.js
+++ b/app/components/workflow-repositories.js
@@ -80,7 +80,13 @@ export default Component.extend({
     if (repos.length > 1) {
       repos = [repos[0]];
     }
-    return repos;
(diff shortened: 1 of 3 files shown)


jrmartino:
tested this, policy cards and repositories have the expected behavior

PR #579, opened 2018-06-15
Search madness

jabrah:
#521 
* Autocomplete now shows up to 100 suggestions as user types in the journal picker.
* When valid DOI is input, the journal search uses `Store.query` by requiring ISSN match and optional match on journal name. This prevents the ember app from creating placeholder Journal objects in Fedora when it is not supposed to.
* Data loading for `submissions/new` cleaned up a bit, now that we don't have to pre-load journals.

### DIFF
--- a/app/components/workflow-basics.js
+++ b/app/components/workflow-basics.js
@@ -143,7 +143,7 @@ export default Component.extend({
           }
           this.set('doiInfo', doiInfo);
           // useful console.log
-          console.log(doiInfo);
+          // console.log(doiInfo);
           publication.set('title', doiInfo.title);
 
           publication.set('submittedDate', doiInfo.deposited);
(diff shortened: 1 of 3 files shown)


karenhanson:
One more comment on this. I noticed that, not only do duplicate records get created, but it looks like those records may be permanent? Is this intended? If it is, I think that's a good thing, so long as duplicate records are not created as was the case in my original test. I wasn't sure if they were supposed to be removed at some point since you're calling them "placeholder" journals.

jabrah:
This is the intended behavior. Calling them placeholders I guess was not great on my part :)

I have a separated branch locally the cleans up the DOI looks in an attempt to eliminate the duplicates issue you mention, but I will submit those changes in a different pr

karenhanson:
@jabrah I will merge this one, and look out for the duplicates change

Issue #564, opened 2018-06-14
Display information about the manuscript throughout the submission workflow


karenhanson:
Under or near to the "New Submission" heading, display some simple formatted article metadata that will be visible through each workflow step. Maybe something like...

### New Submission 
_for_ ***Some Title of Some Article Here (Publication date, if known)***

Rytiggy:
## Verification: below is the change that should happen in this ticket. 
### Before 
![screen shot 2018-06-18 at 2 01 58 pm](https://user-images.githubusercontent.com/17414914/41553661-698e37d2-7300-11e8-9e61-d96fd14205c6.png)
### After 
![screen shot 2018-06-18 at 2 02 20 pm](https://user-images.githubusercontent.com/17414914/41553685-76cad4aa-7300-11e8-8ecd-88b7009cac60.png)

htpvu:
@Rytiggy I dont think that will work. Submission title could get verylong. I would leave the `New Submission` header as it was and add the article title as a `sub title`. Try to make it more obvious so user dont miss it.

htpvu:
Assigned to @jabrah so he can italicize the subtitle text.

PR #592, opened 2018-06-18
Add title to submission workflow title

Rytiggy:
https://github.com/OA-PASS/pass-ember/issues/564
Before
screen shot 2018-06-18 at 2 01 58 pm

After
screen shot 2018-06-18 at 2 02 20 pm


htpvu:
@Rytiggy Please see my comment in the ticket https://github.com/OA-PASS/pass-ember/issues/564 - essentially, we cannot display article title as part of the new submission header. article title can get very long. 

Please display it as subtitle instead.

Rytiggy:
screen shot 2018-06-19 at 9 21 31 am

Issue #422, opened 2018-06-01
Exclude weblinked repositories from the submission object repository list


htpvu:

htpvu:
@cameronblandford is this still a question or has this been resolved after we talked the other day?

PR #603, opened 2018-06-19
web-link repos and specifically ERIC should no longer be included in new submissions

jabrah:
#422 

## Testing
* Start up `.docker/shib` stack as normal
* Login to `ed-user` in order to get some grants that linked to funders with repositories that submissions _must_ be done manually - in this case, US Dept of Education's ERIC
* Start a new submission, pick a grant linked to Dept of Edu such as `Testing the Efficacy of A Developmentally-Informed Coping Power Program in Middle Schools`
* Progress through the rest of the submission workflow, open up the ERIC site, then click Submit to create the new submission.
* After a few seconds (10s max), check in the Submissions table for your new submission. JScholarship should be the only repository shown.

### DIFF
--- a/app/controllers/submissions/new.js
+++ b/app/controllers/submissions/new.js
@@ -1,5 +1,6 @@
 import Controller from '@ember/controller';
 import ENV from 'pass-ember/config/environment';
+import { debug } from '@ember/debug';
 
 export default Controller.extend({
   currentUser: Ember.inject.service('current-user'),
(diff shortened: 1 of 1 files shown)


karenhanson:
@jabrah @htpvu So it does behave as described in the test, but I'm not sure if my other feedback is for this or a separate issue. If you uncheck JScholarship, it makes you upload the files anyway, which is odd because they aren't actually submitted anywhere. Then the submission appears in the Submission table with no repository assigned.

Issue #345, opened 2018-05-18
Handle session expiration more gracefully.


htpvu:
figure out a good strategy to deal with expires session

cameronblandford:
@htpvu what behavior are you seeing right now?

htpvu:
humm. i dont remember. Let's chat with Elliot. he experienced it and request this feature. let's talk to him after scrum

htpvu:
I asked Elliot about the behavior, he doesn't remember what the UI behavior is currently, but he noted that requests to the backend from ember would get spiked. we need to either, continuously ping something in the backend to keep the session live, or detect the expiration of the session and inform users of it, giving them a chance to continue the session

htpvu:
The current behavior when session time out, is that the pages would still load with assets missing. The desired behavior is that when the session times out at the time of some user triggered request, he/she be redirected to home page with a message about them being logged out.

jabrah:
Notes for myself: When a Shib session is expired and a given request is denied access, it seems you get a 302 response. I can't seem to intercept this response to act on, instead it is automagically picked up and dumped in the `ember-fedora-adapter#normalizeResponse` function.

PR #615, opened 2018-06-20
Proposed workaround for shib timeout

jabrah:
#345 

`ember-fedora-adapter` has been updated to throw a specific error when Shib responds with a redirect when session times out. This PR adds a global error handler that will force the current page to refresh when this error is encountered.

## Manual testing
_(Kind of optional)_ If you don't do this modification, I think the default session timeout is one hour
* You can modify the `sp` container from `pass-docker` so that the session expires quickly, retag and build the new image.
  * `pass-docker/sp/etc-shibboleth/shibboleth2.xml`: `<Sessions lifetime="5" timeout="5" relayState="ss:mem" checkAddress="false" handlerSSL="false" cookieProps="https">` seems to do OK
  * `pass-docker/sp/etc-httpd/conf/sp.conf`: any references to `app/` redirection must be removed, as in the `pass-ember` environment, the app sits at `pass/` and not `pass/app/`
* Modify `.docker/shib/docker-compose.yml` to point to the `sp` image you just created

-----

* Run `docker-compose up` from `pass-ember/.docker/shib`
* Login as some known user, wait until session times out, then try to navigate to the Grants page
* The page should reload.

### DIFF
--- a/app/initializers/application.js
+++ b/app/initializers/application.js
@@ -1,0 +1,14 @@
+export function initialize(/* application */) {
+  Ember.onerror = (error) => {
+    console.log(` >> Ember error caught: ${error}`);
+    window.location.reload(true);
+  };
+
+  Ember.RSVP.on('error', (error) => {
+    console.log(` >> RSVP error caught: ${error}`);
+  });
+}
+
+export default {
+  initialize
+};
--- a/app/initializers/application.js
+++ b/app/initializers/application.js
@@ -1,11 +1,16 @@
 export function initialize(/* application */) {
   Ember.onerror = (error) => {
     console.log(` >> Ember error caught: ${error}`);
-    window.location.reload(true);
+    if (error.message === 'shib302') {
+      window.location.reload(true);
+    }
   };
 
   Ember.RSVP.on('error', (error) => {
     console.log(` >> RSVP error caught: ${error}`);
+    if (error.message === 'shib302') {
+      window.location.reload(true);
+    }
   });
 }
 


jabrah:
I want to put this on hold due to some behavior observed in `pass-docker`. Will wait for a new update to `ember-fedora-adapter` before removing the WIP tag.

Basically, during the final submission saving, there are still some errant requests being made to something like `fcrepo/[Object object]`, likely from the file upload logic. This will naturally return a 404 error. In `pass-docker` all 404's respond with `index.html`, so the `ember-fedora-adapter` throws the same error originally meant for the Shib redirect.

Possibilities:
* Find and destroy these odd get requests OR
* Shore up `ember-fedora-adapter` to try to differentiate between a Shib redirect and a 404's `index.html` response

jabrah:
Right. PR needs to be updated now that ES sits behind Shib as well

karenhanson:
@jabrah oh I see, this is my use of pass-docker coming into play again.

jabrah:
Hmm, the workaround in `ember-fedora-adapter` should cover `Store.query`. Will have to poke around again

jabrah:
Problem likely comes from: https://github.com/OA-PASS/ember-fedora-adapter/blob/master/app/adapters/fedora-jsonld.js#L249-L257

This function does not see `results` when unauthenticated, so just dies. Would explain these particular JS errors.

jabrah:
Turns out it is indeed the same problem with the response being a stringified HTML document with various shib things in it. Adding the same workaround as before for now.

Issue #609, opened 2018-06-20
Internal links need to be prefixed with {{rootUrl}}


karenhanson:
Internal links are used in several places. These are formatted starting with `/`, for example, `/img/imagename.gif` or `/faq#processing`. The change to use `/app` in the URL broke these internal links. These need to be prefixed with `{{rootUrl}} so that they resolve regardless of the app sub-path.

PR #620, opened 2018-06-20
Prepend {{rootURL}} to all internal paths

karenhanson:
#609


jabrah:
We have to keep in mind that when static pages are separated, we will likely have to revisit these changes. But these will allow the links to work in `pass-docker` right now :)

PR #625, opened 2018-06-21
Removed link that no longer exists

karenhanson:
Article processing charge FAQ was removed. Removing link from welcome page.


karenhanson:
will include this in faq pr

PR #627, opened 2018-06-21
[WIP] Some cleanup for DOI lookups and validation

jabrah:
## Low priority merge.

This was an older branch that never got a PR and doesn't directly relate to an active ticket


jabrah:
PR is actually done, but we probably shouldn't merge it until after launch at some point. Leaving the PR up to keep it on the radar

PR #629, opened 2018-06-21
'container-title' should now be 'journal-title' in metadata blobs

jabrah:
`journal-title` property now appears in the blob with `id="common"`. Journal name is pulled from CrossRef, not from our Fedora data (they may be different).

``` javascript
[
  {
    "id": "common",
    "data": {
      ...
      "journal-title": "Some journal name goes here",
      ...
    }
  },
  ...
]
```


jrmartino:
verified that "journal-title" correctly appears in the metadata0

Issue #588, opened 2018-06-18
The file upload step of the workflow should wait until the file is uploaded before proceeding to the thank you page


JeffSpies:
We need the following analysis:

- Is the upload page not blocking while uploads are in progress?
- Do we need to use an upload library?
- Can we display a spinner/progress-message during file upload?

That could generate one of two tickets:

1. Replace current file upload with X upload library
2. Use completion event to progress in workflow after files are uploaded; add spinner/progress-message.

JeffSpies:
@htpvu do you want to object creation to hire PRIOR to upload of files? Is that the requirement or is the requirement to fix the problem assuming the true flag will be set only when files are uploaded?

htpvu:
@JeffSpies it is a requirement not to have a submission object with `submitted=true` until after the files are uploaded. 

i didn't see that message until just now

htpvu:
moving back to todo to see if we can resolve the problem of submission object having submitted status BEFORE file upload is done.

PR #634, opened 2018-06-21
Strengthen file-uploading

cameronblandford:
This PR contains the addition of two error-catchings and makes sure that files are only saved if they're present. This PR should help users identify errors more easily, and developers debug any future problems they might find more easily. For better or for worse, I was not able to replicate any issues as they were described in ticket #588.


Rytiggy:
LGRTM

Issue #635, opened 2018-06-21
Files with size > ~100mb don't upload properly


cameronblandford:
Instead of the files being uploaded, the tab crashes when attempting to load the entire large file into local memory. This could also be a symptom of having to write the entire file binary to the body of a POST request. Regardless, loading the whole file at once ceases to be an option after a certain size is reached, and due to fedora not (currently?) supporting multi-part file uploads, there's not a lot that can be done on the front-end side.

birkland:
Do browsers have to load it in memory?  Streaming (transfer encoding
chunked) would work.  It looked like there were some newer
streaming-friendly APIs in one of the various js specs (reading/writing as
a stream rather than byte array)

htpvu:
what is the actually cut off in size? it's an extreme edge case for a manuscript file or supplement file to be larger than 100mb. 

at this point we will need to just put a file size limit notification on the file upload page.

birkland:
It looks like the code reads the contents of the file into a byte array, so
it is browser and memory dependent.  Fedora has no limit.  Maybe we just
say 100, then fix the upload implementation later.

cameronblandford:
@htpvu it's somewhere between 115MB and 140MB. I think 100MB is a good clean limit that's not cutting it too close. File size validation and some text explaining the file size limit wouldn't be much work.

cameronblandford:
@birkland didn't mean to imply that fedora had a hard limit, i wrongly assumed there'd be some sort of soft or hard limit to the size of a post request (just did some quick research and i was wrong, it's a configurable setting on the server-side)

cameronblandford:
I imagine it's a browser memory (or maybe more specifically tab memory on chrome?) issue for sure.

htpvu:
@cameronblandford could you please implement that for us (size validation and proper messaging)?

PR #636, opened 2018-06-21
Add file size validation

cameronblandford:
Add a note that tells users that files must be smaller than 100MB. If a user tries to upload a file larger than 100MB, alert the user that they have done so, in addition to telling them the name of the too-big file and its size.

* #635

### DIFF
--- a/app/components/workflow-files.js
+++ b/app/components/workflow-files.js
@@ -45,6 +45,10 @@ export default Component.extend({
           for (let i = 0; i < uploads.files.length; i++) {
             const file = uploads.files[i];
             if (file) {
+              if (file.size > (1024 * 1024 * 100)) {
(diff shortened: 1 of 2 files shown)


app/components/workflow-files.js
```diff
@@ -45,6 +45,10 @@ export default Component.extend({
           for (let i = 0; i < uploads.files.length; i++) {
             const file = uploads.files[i];
             if (file) {
+              if (file.size > (1024 * 1024 * 100)) {
+                toastr.error(`Your file '${file.name}' is bigger than 100MB (${Number.parseFloat(file.size / 1024 / 1024).toPrecision(3)}MB) and was not added to the submission.`);
+                continue;
```
karenhanson: I found the current file size being in parentheses after the limit slightly confusing :
"Your file 'testfile3.txt' is 113MB. This is larger than the maximum upload size of 100MB and the file was not added to the submission."

PR #640, opened 2018-06-22
Change file size message

karenhanson:


cameronblandford:
Made the same commit myself and was about to make a PR for it :)

Issue #568, opened 2018-06-14
Change how status is displayed for submissions in the Submissions table


jabrah:
Status column should be changed to be calculated based on the combined statuses of all `repositoryCopy`s associated with the submission

* No `repositoryCopy` objects associated with the submission: display nothing
* `submission.source="other" && submission.submitted=false`: display `Manuscript expected`
* All `repositoryCopy` objects have `copyStatus=complete`: display status `complete`
* Other combination of `repositoryCopy.copyStatus`: display a link saying "See details" that links to the submission details page. This part relates to #539 which will have a nicer way of displaying all `repositoryCopy.copyStatus`

-----

**Updated specs: https://github.com/OA-PASS/pass-ember/issues/568#issuecomment-397420573**

karenhanson:
@jabrah I believe status display is still undecided unless you recently talked to @htpvu. The problem is that we may not have a RepositoryCopy for every Deposit. Some Deposits may be accepted by the target Repository and that's the last we know of them.

emetsger:
@jabrah @karenhanson @htpvu yes this probably deserves larger discussion.

## Concrete example: NIH submissions 

1. Deposit Services processes the `Submission`
2. Upon successful processing this results in a `Deposit` with a status of `SUBMITTED`. 
    * A `tar.gz` is dropped off at the NIH FTP server.
    * All we know at that time is whether or not the bytes of the package were transferred successfully
    * There is no indication of acceptance of custody from NIH at the time the package is created and dropped off on the FTP server.
    * In the absence of an acceptance of custody, we ought to leave the status of the `Deposit` as `SUBMITTED`.
3. No `RepositoryCopy` is created.
    * Because there is no acceptance of custody, there's no location for a `RepositoryCopy` to reference, so at the moment Deposit Services does not create one.

After some time in the future, the NIH will process the `.tar.gz` file we dropped off, and they send an email that indicates the acceptance (or not) of custody.

After that email is received, we can change the status of the `Deposit` to `ACCEPTED` (or `REJECTED`, if they do not take custody).  If the email indicates a location of the content, we can create a `RepositoryCopy`.

htpvu:
FOR SUBMISSION TABLE: 

Status column should be changed to be calculated based on the combined statuses of all `repositoryCopys` associated with the submission
* `submission.source="pass"` && `submission.submitted=true` && no `repoCopy`: display `submitted`
[...]

Issue #539, opened 2018-06-12
Consolidate information about repository, deposit and repoCopy in a view that make sense to user


htpvu:
In submission details page

Details to come

jabrah:
Original Submission Details page:
![screenshot from 2018-06-14 10-33-37](https://user-images.githubusercontent.com/5167587/41418796-80b0f19c-6fbe-11e8-9200-3a24f0386f4b.png)

jabrah:
First draft at consolidating repositories, repositoryCopies, deposits:
![screenshot from 2018-06-14 10-33-15](https://user-images.githubusercontent.com/5167587/41418930-d686ab5c-6fbe-11e8-932d-c718254bb29c.png)

jabrah:
Each `repositoryCopy` object related to the submission should be displayed and organized by repository. The status for each should be inferred from the `repositoryCopy.copyStatus` like:
*  `accepted|in-progress` display as "in progress"
* `stalled` as "stalled"?
* `complete` as "complete"

htpvu:
We need to account for deposit that do not have or have not had RepoCopy objects. So the information needs to start out with the list of repositories attached to the submission object:
For each repository:

- display repository name
- display status:
-- RepoCopy exists with status `complete`: `complete` (tooltip: Submission to this repository was accepted and processed that the repository. ID(s) have been assigned to the submitted manuscript)
-- RepoCopy exists with status `stalled`: `stalled` (tooltip: We detected that the target repository has stalled processing of your manuscript. Please sign on to the target repository's system for more details)
-- RepoCopy does not exist: `submitted` (tooltip: PASS has received your submission. Your submission has been sent to the target repositories or is in queue to be sent)
- display repoCopy's external Ids if they exist.

jabrah:
Add status check:
* `repoCopy exists && submission.source == 'other' && submission.submitted == false` then display status `Manuscript expected` with appropriate tooltip

htpvu:
THis looks rather beautiful!

PR #648, opened 2018-06-22
Update for submission status

markpatton:
The submissions table, submissions detail page, and grants detail page have their logic for calculating submission status updated. In addition the submissions table will make a constant number of queries instead of being linear to the number of rows.

Due to limited time, there is code duplication in submissions/index and grants/detail routes as well as in the
submissions-status and submissions-status-cell components.

In order to test. Fire up the app login as various users and look at the submission status.

Note that I am not sure the logic is correct!

### DIFF
--- a/app/components/submission-status-cell.js
+++ b/app/components/submission-status-cell.js
@@ -1,32 +1,14 @@
 import Component from '@ember/component';
+import { inject as service } from '@ember/service';
 
 export default Component.extend({
-  // For specs: https://github.com/OA-PASS/pass-ember/issues/568
+  submissionStatus: service(),
+
   status: Ember.computed('record', function () {
     const submission = this.get('record');
     const repoCopies = this.get('column.repoCopiesMap')[submission.get('id')];
     const deposits = this.get('column.depositsMap')[submission.get('id')];
 
-    if (repoCopies.any(rc => rc.get('copyStatus') == 'stalled')) {
-        return 'Stalled';
-    }
-
-    if (submission.get('source') == 'other' && !submission.get('submitted')) {
-        return 'Manuscript expected';
-    }
-
-    if (repoCopies.get('length') < deposits.get('length')) {
-      if (submission.get('source') == 'pass') {
-        return 'Submitted';
-      }
-    }
-
-    if (repoCopies.get('length') > 0 && repoCopies.get('length') == deposits.get('length')) {
-      if (repoCopies.every(rc => rc.get('copyStatus') == 'complete')) {
-        return 'Complete';
-      }
-    }
-
-    return 'See details';
+    return this.get('submissionStatus').calculateStatus(submission, repoCopies, deposits);
   })
 });
--- a/app/services/submission-status.js
+++ b/app/services/submission-status.js
@@ -1,0 +1,37 @@
+import Service from '@ember/service';
+
+export default Service.extend({
+
+  // Determine the status of a submission given associated repoCopies and deposits.
+  // TODO Document the status semantics and unit test.
+  // Some (out of date?) information https://github.com/OA-PASS/pass-ember/issues/568
+  calculateStatus(submission, repoCopies, deposits) {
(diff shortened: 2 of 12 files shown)


app/templates/components/submission-status.hbs
```diff
@@ -0,0 +1,46 @@
...
+    {{#if (eq status "Submitted")}}
+      <span tooltip-position="left" tooltip="This submission has either been sent to all target repositories, or is in the queue and will be submitted shortly.">
+        <i class="fas fa-info-circle"></i> {{status}}
+      </span>
+    {{/if}}
+    {{#if (eq status "Stalled")}}
+      <span tooltip-position="left" tooltip="Manual intervention required.">
+        <i class="fas fa-info-circle"></i> {{status}}
+      </span>
+    {{/if}}
```
karenhanson: Per the description here, this should be `See details`, there is no stalled status show for a combined status calculations, so this should not be displayed on the Submissions table or at the top of Submission Details page

https://github.com/OA-PASS/pass-ember/issues/568

> * submission.source="pass" && submission.submitted=true && no repoCopy: display submitted
> * submission.source="other" && submission.submitted=false && no repoCopy : display Manuscript expected
> * ALL of repoCopy.copyStatus="complete": display complete
> * Other combination of repositoryCopy.copyStatus: display a link saying "See details" that links to the submission details page. This part relates to #539 which will have a nicer way of displaying all repositoryCopy.copyStatus

karenhanson: To clarify - the one at the top should say "see details" (is currently incorrect), the one on the repository should say "stalled" (is currently correct)

image

app/components/submission-status-cell.js
```diff
@@ -1,43 +1,32 @@
...
-      if (source == 'pass' && submission.get('submitted')) {
-        return 'Submitted';
-      } else if (source == 'other' && !submission.get('submitted')) {
+    if (submission.get('source') == 'other' && !submission.get('submitted')) {
         return 'Manuscript expected';
+    }
+
+    if (repoCopies.get('length') < deposits.get('length')) {
+      if (submission.get('source') == 'pass') {
+        return 'Submitted';
```
karenhanson: Submitted should appear when there are zero RepositoryCopies for the Submission, `Submission.source=pass` and `Submission.submitted=true`. Currently these are appearing as more details links.

markpatton:
The status logic really needs to be pulled out into a service where we can unit test the various permutations.

markpatton:
I pushed the logic into a service and updated it as requested. I think. This needs a full specification and unit testing to really be confident about.

I won't have any time to work on this for a few days!

htpvu:
I'll stew on this over the weekend.

karenhanson:
I think because of a minor lag in creating the deposit and indexing, the query fails if you go right to the submission list from the Thank You page. There are supposed to be ~15 more submissions below this. Refreshing fixes it. So perhaps  the default "see details" can be shown if no records are found and `source=pass`?

image

app/services/submission-status.js
```diff
@@ -0,0 +1,37 @@
...
+    if (repoCopies.get('length') > 0 && repoCopies.get('length') == deposits.get('length')) {
+      if (repoCopies.every(rc => rc.get('copyStatus') == 'complete')) {
+        return 'Complete';
+      }
+    }
+
+    if (repoCopies.get('length') > deposits.get('length')) {
+      if (submission.get('source') == 'other' && submission.get('submitted')) {
+        return 'Complete';
+      }
```
karenhanson: So this isn't exactly correct - you can have an in-progress submission added through NIHMS ETL. The repository Copy status in that instance would be "accepted" or "in-progress"... so in order to be complete you still need this logic: `if (repoCopies.every(rc => rc.get('copyStatus') == 'complete')) {`

karenhanson:
Merged though there are still a few issues so we can deal with those on a separate PR

Issue #646, opened 2018-06-22
2. Grants - clicking on award number in grants step takes you out of workflow


karenhanson:
You can click on an award number in `2. Grants`, but when you go back you have lost your place. Any links out of the workflow need to be `target="_blank"` or removed. Note that the "contact us" link, also on Grants step does the same thing.

Issue #584, opened 2018-06-17
[WIP]Repositories are not being attached to Submission object at creation!!!!!!


htpvu:
on rev. 96d1830b

I verified (through the view of ember inspector data) that the submission object had an empty respositories list. This submission should have had JScholarship in it. 

I looked at the resources in the repo, this submission object had NO repositories attached to it

 <img width="1317" alt="screen shot 2018-06-16 at 10 20 53 pm" src="https://user-images.githubusercontent.com/11050142/41504137-f0668640-71b3-11e8-87fa-fbe941a0f76c.png">

htpvu:
need more investigation

Rytiggy:
@htpvu I have not been able to replicate this. Could you provide more specific steps to reproduce the bug?

karenhanson:
@Rytiggy @htpvu I know of 3 scenarios where you can get an empty repository list. 

Log in as `ed-user`, pick a Department of Education grant, then either (1) de-select JScholarship and only submit to a linked repository (e.g. ERIC), or (2) don't check the agreement so it removes JScholarship, then the repository list is empty. The 3rd scenario is if you only have JScholarship and do not check the agreement but proceed anyway, it will take you all the way to the end of the workflow and you will get a submission with no repositories. I have not seen a case where you submit to JScholarship correctly and the repository list is empty - is it possible that you did not check the agreement and it completed anyway?

htpvu:
@Rytiggy @karenhanson - i think what you describe karen are real scenarios created by the gap in code logic. what i observed were in the implemented code paths, but something were misbehaving. I haven't been able to reproduce the original bug either.

In regard to the 3 scenarios that youdescribed, at least one should have been address by previous work by @cameronblandford - i'm not sure what happen to the popup asking user to leave the workflow. so we shold look into these thigns.

PR #649, opened 2018-06-22
Feature/add alert when no repos are added

Rytiggy:
https://github.com/OA-PASS/pass-ember/issues/584
https://github.com/OA-PASS/pass-ember/issues/646
add redirect to `/logout` on press of logout button


karenhanson:
@Rytiggy I'm merging this, I will fix the last link for you. Message me if you already did it.

Issue #655, opened 2018-06-24
Perform journal metadata look up in the case submission creation without DOi


htpvu:
Copy of Slack:

the usecase is this: i’m a user, without DOI. i typed in my journal article title, then selects from the journal lists my journals
the package sent away did not include any journal information
What needs to happen in this case is that when user *SELECT* from the journal list, we should retrieve the ISSN of that journal from our database
and perform the regular look up
so that the required information can be included in to the BLOB as well as the deposit

PR #656, opened 2018-06-24
Manually add metadata to submission when not using a DOI

jabrah:
#655 

* User can now enter author info in the metadata step
* Manually entered article title and journal data is added to the metadata blob
* Manually entered article title and journal info is prepopulated into the metadata step

Addition:
* `issn-map` should not appear in either the submission review _or_ submission details pages

## Testing
* Load up docker and login as normal
* Start a new submission
* ### **Do not enter a DOI** 
* Enter stuff for _article title_ and select a journal. Click Next
* Progress until you get to the Details step. 
  * Article title, Journal ISSN should appear correctly
  * User should be able to add authors here (hint: add a user)
* Progress with the rest of the submission
  * **Make sure you have your developer tools open for you browser, opened to the Network tab**. This way you can monitor the outbound requests
* Click the final Submit button
  * In the dev tools, you will see a `submissions` request. The response of this will have the new submission ID in Fedora. You can directly navigate to this URL to see the submission object in Fedora
* Bring up the specific submission in Fedora and inspect it's `metadata` property - this will be a stringified JSON blob. 
  * Article title, journal name, issn, and issn-map should all be present
  * If you had added an author, that author should also be present

### DIFF
--- a/app/components/workflow-basics.js
+++ b/app/components/workflow-basics.js
@@ -88,6 +88,9 @@ export default Component.extend({
       if (!this.get('doiInfo.title')) {
         this.set('doiInfo.title', this.get('model.publication.title'));
       }
+      if (!this.get('doiInfo.author')) {
+        this.set('doiInfo.author', []);
+      }
       this.sendAction('next');
     },
     validateDOI() {
--- a/app/components/workflow-basics.js
+++ b/app/components/workflow-basics.js
@@ -85,11 +85,8 @@ export default Component.extend({
       }
     },
     next() {
-      if (this.get('doiInfo').length === 0) {
-        this.set('doiInfo', {
-          'container-title': this.get('model.publication.journal.journalName'),
-          title: this.get('model.publication.title')
-        });
+      if (!this.get('doiInfo.title')) {
+        this.set('doiInfo.title', this.get('model.publication.title'));
       }
       this.sendAction('next');
     },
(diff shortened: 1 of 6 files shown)


app/components/workflow-basics.js
```diff
@@ -197,7 +197,20 @@ export default Component.extend({
...
+      let doiInfo = this.get('doiInfo');
+      doiInfo = {
+        'journal-title': journal.get('journalName'),
+        ISSN: journal.get('issns')
+      };
+
+      const nlmtaDump = await this.getNlmtaFromIssn(doiInfo);
+      if (nlmtaDump) {
+        doiInfo.nlmta = nlmtaDump.nlmta;
+        doiInfo['issn-map'] = nlmtaDump.map;
```
Rytiggy: This is getting shown to the user as `[object, object]` should be stringified or not displayed. 
screen shot 2018-06-25 at 12 07 43 pm

Rytiggy: This will also need to be updated in the submission/detail route

Rytiggy: After **those fixs** this PR is good to go

karenhanson: my vote is don't show it! we already have ISSN showing

jrmartino:
followed the above recipe, behavior was as expected

PR #657, opened 2018-06-24
Edited submission status tooltip text, fixed CSS for Stalled

karenhanson:
* Changed tooltip text for stalled
* Changed color of button to red for stalled, fixed css so that this color change works
* Moved the `<style>` tags to the top of the Submissions page as they were being repeated for every cell (that means any page that puts the status cell on will need those style tags, is this OK? Alternatively could move it to main stylesheet)

### DIFF
--- a/app/components/submission-repo-details.js
+++ b/app/components/submission-repo-details.js
@@ -32,7 +32,7 @@ export default Component.extend({
       case 'Complete':
         return 'Submission was accepted and processed by the repository. ID(s) have been assigned to the submitted manuscript.';
       case 'Stalled':
-        return 'The repository has found a problem with your submission that has caused progress to stall. This may require direct interaction with the repository to re-initiate the process. Please check the repository website for more details.';
+        return 'The repository has found a problem with your submission that has caused progress to stall. Please contact the repository for more details.';
       case 'Submitted':
         return 'Your submission has been sent to the repository or is in queue to be sent.';
       case 'Manuscript expected':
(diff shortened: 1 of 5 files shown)


jrmartino:
In order to verify behavior we need to have a submission with stalled status - this would be the status on the RepositoryCopy for the submission. Steps to do this are as follows:

- Open developer tools in chrome. select network tab, Go to submisions table and select a submission. 
Look at "es" elements, select the one for repositoryCopy - header will will look like this:
...
{size: 100, from: 0, query: {bool: {must: {term: {,…}}, filter: {term: {@type: "RepositoryCopy"}}}},…}
from
:
0
query
:
{bool: {must: {term: {,…}}, filter: {term: {@type: "RepositoryCopy"}}}}
size
:
100
_source
:
{excludes: "*_suggest"}

- Open preview tab in devolper tools - if the number of hits is 0, find a different submission
If  number of hits is positive, drill down through "hits" until you see the fedora id for the repo copy. Take this URI and open it in fedora.
- use the SPARQL Update Properties facility there to update the copyStatus - something like this (the repoCopy id I used was https://pass/fcrepo/rest/repositoryCopies/4f/22/c2/70/4f22c270-c34a-41ac-8d77-9ee6e8fb95a6


PREFIX pass: <http://oapass.org/ns/pass#>

DELETE DATA
{<https://pass/fcrepo/rest/repositoryCopies/4f/22/c2/70/4f22c270-c34a-41ac-8d77-9ee6e8fb95a6> pass:copyStatus "complete"};

INSERT DATA {
<https://pass/fcrepo/rest/repositoryCopies/4f/22/c2/70/4f22c270-c34a-41ac-8d77-9ee6e8fb95a6> pass:copyStatus "stalled"};

Issue #681, opened 2018-07-10
Enable support for pseudo-super-user 


htpvu:
a super user can see all things via the UI, and can submit against all grants or without grant. 

in absence of a formal user role. a super-user can be supported by having a user carry 2 roles (admin, which allows for global view access and submitter, which allows for submit access) 

UI code need to be updated to check for these two roles where appropriate and ensure person with 2 roles can be indeed see all things and make submission

PR #682, opened 2018-07-10
[do not merge] Add superuser stuff

jabrah:
#681 

* Add checks for `admin` vs `submitter` in various routes
  * `admin` is allowed to see all data (grants and submissions)
  * `submitter` is allowed to see only their data
  * `submitter` is allowed to enter the submission workflow
  * If a user is both `admin` and `submitter` they should be able to see all data AND create submissions

Until paging is implemented, "all" data just means 500 records :(

## Testing
`pass-docker` already has appropriate test data. 

* Edit the `pass-docker` `.env` file
```
  EMBER_GIT_REPO=https://github.com/jabrah/pass-ember
  EMBER_GIT_BRANCH=8537e590d896dcd4a85a2f9c9c69e8eb1e7b0368
```
* Pull appropriate images
* Build the new Ember image: `docker-compose build ember`
* `docker-compose up -d`
* Do testing things
  * Login as `admin-submitter` to test a user with both roles

### DIFF
--- a/app/controllers/dashboard.js
+++ b/app/controllers/dashboard.js
@@ -2,7 +2,5 @@ import Controller from '@ember/controller';
 import { computed } from '@ember/object';
 
 export default Controller.extend({
-  currentUser: Ember.inject.service('current-user'),
-  isSubmitter: Ember.computed('currentUser', () =>
-    this.get('currentUser.user.roles').includes('submitter'))
+  currentUser: Ember.inject.service('current-user')
 });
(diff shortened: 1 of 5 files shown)


karenhanson:
Tested `admin`. It has the same issues as `admin-submitter` but also:
* When a manuscript is expected the "Finish Submission" button appears though it is supposed to be readonly.

jabrah:
My first thought is that most of the above bullet points are tied to the fact that we do not have server side paging implemented. At the moment, there is a maximum number of search results hard coded into the app for any and all searches of 500. Because of this, there is no way for the UI to get grants outside of the 500. This also means that Ember Data can't do its aggressive preloading of data the way it wants to , so occasionally it has trouble resolving relationships.

Do you get similar behavior when testing with the usual users (of role `submitter` only?)

karenhanson:
Functionality went in a different direction. Closing this for now

PR #710, opened 2018-08-06
Add more tests

cameronblandford:
This PR adds a base file for integration tests for each component - since almost all the app's logic happens in components, it made sense to focus here for my last few hours. Most will fail out of the box - to better understand how to write integration tests, turn to either online documentation or the already-existing integration test files for pass-ember (or contact me, and I'll be happy to explain off the clock!).

By getting the tests to pass in their current state (by generating and passing in the correct parameters to each component), you'll have rendered a working version of that component that you can use to create further tests and assertions for that module.

This PR also fleshes out some of the service unit-tests.


app/controllers/application.js
```diff
@@ -14,9 +14,4 @@ export default Controller.extend({
   didRender() {
     this.set('institution', this.store.find('institution'));
   },
```
karenhanson: Looks odd that this finishes on a comma - I realize it did before so I'm going to merge anyway, but wondering if this is correct!

karenhanson:
Looks like there is more work to be done, but will merge the progress since it is primarily tests and removal of a few comments. Just one comment about application.js ending on a comma. Looks odd, but it was like that before and wasn't breaking so I'm going to merge this as is.

PR #706, opened 2018-08-03
Session expiration handling, NLMTA lookup error catching, security vulnerability patching

cameronblandford:
* #345 
* #664 
* #688 
* #689

Issue #689, opened 2018-07-16
NLMTA API lookup does not protect against failure to access service


karenhanson:
If no matches are found when you look up an ISSN, the ember app handles things fine. If, however, the NLMTA API website is inaccessible (I simulated this by changing the domain name for the API call), or the `esummary` page for the journal returns an error (I simulated this by putting an incorrect parameter in the query), the journal lookup errors and you cannot select a journal even if it exists.

PR #712, opened 2018-08-09
Move NLMTA lookup fail to fetch requests

karenhanson:
Extension to PR #706, relates to issue #689. 

Moved error handling to `fetch()` request for NLMTA to (a) ensure the error handling occurs on all calls to the service and (b) ensure that the code logs the issue in the console and continues using an empty return.

### Testing
To test, point to the commit for this PR, go to the first step of the workflow page and complete the form (with and without DOI). Everything should work as usual and the NLMTA code should be loaded if possible. Good, previous functionality is not broken!

The challenge is forcing a service failure to test error handling. To do this for testing, I have created two copies of this branch with slight modifications. The first branch breaks the URL for NCBI to simulate the service being unavailable. Point `pass-docker` to the most recent commit here to see what happens when the service is unavailable:
https://github.com/karenhanson/pass-ember/tree/nlmta-fail-fix-test-noservice 
The other test simulates a bad request that returns something not structured as expected. For this, point docker to this `pass-ember` branch:
https://github.com/karenhanson/pass-ember/tree/nlmta-fail-fix-test-badrequest

In a successful test for these two scenarios, the user should not notice anything unusual, except that no NLMTA code will be displayed. If you open the developer tools, the console log should show an NLMTA Lookup failure.


htpvu:
@karenhanson i've tested both of the above branches. the `nlmta-fail-fix-test-badrequest` branch worked as expected. the `nlmta-fail-fix-test-noservice` only works if i open the dev tools. without dev tool, i'm not able to select a journal from the dropdown list and thus is stuck in the workflow.

PR #716, opened 2018-08-14
Make web linked repositories open correctly in demo

### LINKED ISSUE #714: Clicking on USAID or Education link at the review step does NOT open the link in the new tab
This behavior is observed in the demo instance. It's hard to verify behavior in PROD, but i'm assuming that it's the same. 

To replicate:
- log in to PASS.
- start a new submission against a USAID grant or an ED grant
- go through to the review step
- click on the web portal link
- observe that nothing happens when link is clicked.

PASS UI need to open a new tab/browser loading the link, when a link is clicked by user. 

See screenshot for the link mentioned. 
screen shot 2018-08-13 at 9 15 38 pm

karenhanson:
PR to _hopefully_ fix #714, or at least not break it more!  This issue only appeared in demo and presumably production. It was not replicated in `pass-docker` and will therefore only be fully verified by testing in demo.

The main identifiable and relevant difference between running this on `pass-docker` vs running on demo is that demo URLs have periods in them, while `pass-docker` runs at `https://pass`.  In the Ember code, periods can result in attempts to bind objects as if it is dot-notation. 

The change in this PR alters the current approach to updating the `externalRepoMap`. Curiously, there was already a line commented out that will hopefully fix the problem, so this change switches to use that. I have verified that using this other line of code does not break anything.

### Testing
Using `pass-docker` the branch should be tested to ensure links open as expected when using the `https://pass` URL i.e. the change doesn't break anything. Note that user `usaid-user` has both an education grant and a USAID grant to test adding two web-linked repositories simultaneously.

Once approved, we need to push a new image to the Docker hub, and test the branch in demo to see if the issue when using URLs with periods in them is now resolved.

### DIFF
--- a/app/components/workflow-review.js
+++ b/app/components/workflow-review.js
@@ -96,8 +96,7 @@ export default WorkflowComponent.extend({
         }
         // Go to the weblink repo
 
-        // this.get('externalRepoMap')[repo.get('id')] = true;
-        this.set(`externalRepoMap.${repo.id}`, true);
+        this.get('externalRepoMap')[repo.get('id')] = true;
         const allLinksVisited = Object.values(this.get('externalRepoMap')).every(val => val === true);
         if (allLinksVisited) {
           this.set('hasVisitedWeblink', true);


birkland:
Can it be tested locally by adding something like `pass.local` as 127.0.0.1 to the hosts file, then changing the relevant `.env` variables to use `https://pass.local` instead of `https://pass`?

htpvu:
@karenhanson is this fix expected to work for the Dept of Ed links as well? or JUST USAID?

karenhanson:
@htpvu it should fix both.

@birkland yes it could be tested that way. I will look into that.

karenhanson:
@birkland I think it does appear on the details page but not in the `/submissions/` list if I remember rightly. I can't remember the justification, but I think there was one! :)

htpvu:
@birkland We only provide a link for USAID and ED submission portal, we make sure that they click on such link but has no way of telling weather they did a submission or finish it. so we dont show those repositories as suggestion, but not as something we can track.

birkland:
The behavior of web link submissions still does not entirely make sense to
me, but if the behavior is as expected, then I suppose there are no
additional issues to address.  In any case, the important thing that the
link opening part that pertains to this PR does work, so this PR looks good.

PR #723, opened 2018-08-25
Added function to scroll to top when navbars are rendered

### LINKED ISSUE #719: UI opens submissions table page at mid page
I observed this when Sayeed was demoing at NIH. He observed the same thing when giving another demo this week.

Behavior:
When there are a lot of submissions to be view (I dont know the exact number), the submissions page and the grant details page will open at mid page. Somehow once it's loaded you'll be lookinng at some wher ein the middle of the page and not the top of the page. It was confusing enough for sayeed. 

we need to make sure pages always load at the top.

karenhanson:
This fixes #719.  A script now runs when the workflow navigation bar renders or the top navigation bar renders (which should happen on each new page or workflow step) to set the scroll position to the top of the page.  When going through the workflow or scrolling down the submission page then changing page the user is always returned to the top of the page. This does not affect the scroll when there is a hash in the URL (as with FAQ links).

@htpvu 2 things I wanted to mention. This seems to work, but being unfamiliar with ember I don't know if there is a better way code-wise. Also, note that because the navigation bar tends to load first, sometimes there is a lag in the page content changing when flicking between e.g. grants and submissions - let me know if you think this is unacceptable. There may be a way to improve this.

### DIFF
--- a/app/components/nav-bar.js
+++ b/app/components/nav-bar.js
@@ -17,7 +17,7 @@ export default Component.extend({
   },
   didRender() {
     this._super(...arguments);
-    if (!window.location.hash) {
+    if (window.location.search.indexOf("anchor=") == -1) {
       window.scrollTo(0,0); 
     }
   }
--- a/app/components/nav-bar.js
+++ b/app/components/nav-bar.js
@@ -15,4 +15,10 @@ export default Component.extend({
       // this.get('session').invalidate();
     },
   },
+  didRender() {
+    this._super(...arguments);
+    if (!window.location.hash) {
+      window.scrollTo(0,0); 
+    }
+  }
 });
--- a/app/components/submission-nav.js
+++ b/app/components/submission-nav.js
@@ -14,4 +14,8 @@ export default Component.extend({
       }
     }
   },
+  didRender() {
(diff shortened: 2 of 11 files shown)


htpvu:
I couldn't verify the issue before the fix and couldn't tell the difference AFTER the fix. But didn't encounter any issue while testing, so will merge and push to demo so Sayeed could verify.

karenhanson:
@htpvu Yep, that's not the expected behavior for faq! I will fix that.

If you want to see before/after behavior on e.g. submissions-> grants, in demo... 
(1) Go to Submissions summary, make window small if need a vertical scrollbar to appear. Scroll down
(2) Click for Grants page, it will load mid page.
(3) Start new submission, fill in first step
(4) make window small for vertical scrollbar to appear. Hit next, it loads mid-page

If you do these steps in the new version it should go back to top (except for thing with navbar loading first mentioned in original PR comment, and the newly discovered faq issue).

karenhanson:
@htpvu added fix for anchor tags. There may be an easier way to do this, but this seems to fix both the scroll back to top issue and the problem with using `#` in the URL.

PR #785, opened 2018-11-08
Fix handling of weblink repositories

markpatton:
Changes:
- Fixes some string literal lint issues
- Use Repository.integrationType to find weblink repos
- Updates to assets image with Repository.integrationType
- Removes or comments out various debugger statements
- When creating a proxy submission
  - Do not remove weblink repositories from Submission 
  - Do not add external-submissions metadata to Submission
  - Are not required to click for weblink repositories
  - Submissions details page of new submission
    - Repositories section includes ERIC
    - No link to ERIC provided
    - No "external submission repositories" section
- When approving a proxy submission
  - Remove weblink repositories from Submission
  - Add external-submissions metadata to Submission
  - Are required to click for weblink repositories
  - Submissions details page before approval
    - Repositories section includes ERIC
    - Link to ERIC provided
    - No "external submission repositories" section
  - Submissions details page after approval
    - Repositories section does not include ERIC
    - No Link to ERIC provided
    - External submission repositories section includes ERIC
When submitting on your own behalf to grant that includes ERIC
  - Submissions details page after approval
    - Repositories section does not include ERIC
    - No Link to ERIC provided
    - External submission repositories section includes ERIC

In order to test:
1. Login as nih-user
1. Do proxy submission on behalf of Katrina Debnam
    - Associate with ERIC grant
1. Verify that:
    - Submission contains the ERIC repository
    - Submission does not have external-submissions metadata
    - No click on ERIC link was required
    - The submissions detail page is as above
1. Login as ed-user
1. Verify the submissions detail page is as above
1. Approve the submission
1. Verify that
    - Submission does not contain the ERIC repository
    - Submission does contain a external-submissions attribute containing information about the ERIC repository
    - Click on ERIC link was required
    - The submissions detail page after submission is as above

Closes #783 #775

PR #786, opened 2018-11-08
Remove URI encoding hack

markpatton:
This pr removes a hack in the grants and submissions detail page which checked for an unencoded grant or submission URI and encoded it.

The tests fail because of string template issues. But this is fixed in #785 .


birkland:
Note:  It is possible that this was a workaround for decoding happening at the Shibboleth or load-balancer level.  Testing in AWS is necessary.

markpatton:
Hack is still needed.

PR #794, opened 2018-11-09
Encoded uris

### LINKED ISSUE #787: Update ember's hack to better encode and unencode URIs it is given
ember encodes everything behind following `/submissions/`, which it should not. need to be updated to not encode anything after `?`

karenhanson:
There is an issue with passing params, this lops off params rather than tries to add them again
resolves #787. 

You can test by passing in the unencoded version of the link with parameters after. For example, before the fix, links like this would have caused files not to load:
* https://pass.local/app/submissions/https://pass.local/fcrepo/rest/submissions/cc/f7/55/87/ccf75587-d828-4458-9deb-d1ed108d3896?userToken=BRO7QA373FK4LPNEJN3TRZUTO2B24EPQ6BHUJOOX3ZOWZEYMXFYRHMPK35THFHHTN4BCLRCXRIKEOX3AOPWCDIRLRCA6DK3LJXGQTSUKWAYZP23IWVMWF5IHQECOMNMVLHOKJTG2LZWNH7DCTZJZPOTYKM6HDJX65TE53ZGFNBYHGOMHPJB3XZET5HPOTFBQ52LHFJQJM24VQCVVNBVRXR5B3FDCATSYRVR7A2CXDP233VFOAIBAOW6LKBTIMGS7CHM4E3U7RHD5H4W2QQX5GUU5JQ
* https://pass.local/app/submissions/https://pass.local/fcrepo/rest/submissions/cc/f7/55/87/ccf75587-d828-4458-9deb-d1ed108d3896?foo=bar 

Now, these should load the details page fine, but you will note that the parameters are no longer visible in the URL.

### DIFF
--- a/app/components/workflow-basics.js
+++ b/app/components/workflow-basics.js
@@ -251,7 +251,7 @@ export default WorkflowComponent.extend({
     validateEmail() {
       const email = this.get('submitterEmail');
       let emailPattern = /^[a-zA-Z0-9._-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,4}$/;
-      if (email && emailPattern.test(email)) { 
+      if (email && emailPattern.test(email)) {
         this.set('validEmail', 'is-valid');
       } else if (email) {
         this.set('validEmail', 'is-invalid');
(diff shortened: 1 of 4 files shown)


markpatton:
This seems to work. The application route is where userToken is interpreted. The details routes don't matter. But there is a TransitionAborted error caused by the replaceWith to fix the route id. I am not sure if it is innocuous or not.

PR #803, opened 2018-11-12
Avoid stripping out preparers, better proxy checking

markpatton:
This pr comments out a check that strips preparers. It is triggered when an approver tries to edit a submission. It is unclear to me what the check was trying to do.

Other changes include more precise checking for a proxy in the new submission route when handling external-submissions metadata and stripping out web-link repositories.

### DIFF
--- a/app/components/workflow-basics.js
+++ b/app/components/workflow-basics.js
@@ -172,7 +172,7 @@ export default WorkflowComponent.extend({
       const proxySubmitterExists = submitterExists && currentUserIsNotSubmitter;
 
       if (!proxySubmitterInfoExists && !proxySubmitterExists) {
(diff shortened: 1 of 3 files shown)


htpvu:
IN testing this PR, we'll need to test for:

existing functionalities --> make sure they still work, i.e. self submission workflow still works. 
new submissions --> the displays and prompt for weblink repo make sense and does not break things

examine objects as they change state.

karenhanson:
Tested regular submission, prepared->approved, prepared->cancelled, prepared->send for edits->prepared->approved with web link, prepared-> edited by submitter ->submitted. In all cases the preparer and submitter were correct.

PR #816, opened 2018-11-14
Fixes policy page logic to show radio buttons for policies using PMC

### LINKED ISSUE #790: Submission subjected to policies which uses PMC/NIHMS should be prompted about the various journal methods
Currently, only Submission under NIH policies are given a choice to opt out of submitting through PASS by making arrangment with publisher. This choice should be made available to other agencies who uses the PMC infrastructure to support their policies as well.

karenhanson:
The logic that displays the radio button on the policy page now applies to any policy that uses PMC as its repository. This uses the new `repositoryKey` value, and so an updated `assets` image is referenced with that value populated.

Includes a couple of updates to the model:
* adds `Repository.repositoryKey` - this fix needs the assets container that has these values
* removes `User.institutionalId` (I verified that nothing references this)
* adds `User.locatorIds`

closes #790

### DIFF
--- a/.docker/shib/docker-compose.yml
+++ b/.docker/shib/docker-compose.yml
@@ -125,7 +125,7 @@ services:
     depends_on:
       - assets
   assets:
-    image: birkland/assets:2018-11-07_3.2@sha256:5043faf1e43505e45c013e8f0542b7494ddd6379bcf3714607a42f48bd1f4f3e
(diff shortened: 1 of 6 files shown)


jrmartino:
i'm not familiar with the ember code at all, but am commenting that User.institutionalId and User.localKey are both in the same boat - obsoleted by locatorIds.

htpvu:
I did not observe the desired effects. AM i doing it wrong? 
screen shot 2018-11-14 at 10 45 59 pm

karenhanson:
@htpve Actually never mind I see the issue... it's a quick fix.

PR #817, opened 2018-11-14
Use a paginated table to select grants during submission.

### LINKED ISSUE #813: Update the grant selection page in submission workflow to better handle the display and selection of grant from a long grants list
I started a submission on behalf `faculty1`, the grants list for him that showed up in the grant selection step were about 5-6 items long. but when logged in as `faculty1`, he has a grants list of 109 item long. 

Someone please verify this. a bad query?

markpatton:
- Previously a user could select a maximum of 10 grants from a list. 
- Now a paginated table of available results is presented to the user.
- Grants are sorted by descending end date
- The user can click on a row to toggle the grant being associated with the submission.
- The chosen grant list is automatically synced with the table
- Coveralls (used for test coverage check) is turned off in Travis.

The basic functions seem to work. There is probable more to do on styling and layout.

In order to test, try a user with a lot of grants like faculty1, create a new submission, and play with the grants page.

Closes #813

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ before_install:
 install:
 - npm install
 script:
-  - ember test --test-port=4200 && echo "Reporting coveralls" && cat coverage/lcov.info | node_modules/coveralls/bin/coveralls.js && rm -rf ./coverage
+  - ember test --test-port=4200
 before_deploy:
   - ember build --environment=surge
 deploy:
--- a/app/components/select-row-toggle.js
+++ b/app/components/select-row-toggle.js
@@ -1,0 +1,14 @@
+import Component from '@ember/component';
+import layout from '../templates/components/select-row-toggle';
+
+export default Component.extend({
+  layout,
+  actions: {
+    clickOnRow(index, record, event) {
+      console.log('hi there buddy');
+
+      this.get('clickOnRow')(index, record);
+      event.stopPropagation();
+    }
+  }
+});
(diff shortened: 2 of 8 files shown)


htpvu:
This works great.

I do have some stylistic feedback. please ssee screenshots:


screen shot 2018-11-14 at 11 22 50 pm

screen shot 2018-11-14 at 11 33 36 pm

karenhanson:
@markpatton one more minor thing actually... noticing some word wrapping on the funder name, so I was looking at the grants table, which doesn't wrap. Turns out the width is controlled like this: 
https://github.com/OA-PASS/pass-ember/blob/b4b68afeeb6e4325de594ec853b5721748131547/app/styles/app.scss#L520

So, I think it would be good to define a style for the funder column, as is done for award number:
https://github.com/OA-PASS/pass-ember/blob/b4b68afeeb6e4325de594ec853b5721748131547/app/styles/app.scss#L712
https://github.com/OA-PASS/pass-ember/blob/b4b68afeeb6e4325de594ec853b5721748131547/app/controllers/grants/index.js#L45

That way the method used in the first snippet could be eliminated. I can do a PR on your branch if you want me to take it off your plate.

markpatton:
Various fixes made:
- Only show pagination when there are multiple pages
- Set colors on pagination as requested. (Would black arrows look better?)
- Text update as requested
- I tried the color change for the checkboxes, but it looked odd to me in both checked and unchecked forms. I think something higher contrast is needed for such small icons.

markpatton:
@karenhanson

As you suggest we could use a similar styling on funder as award number, but it seems like wrapping is inevitable given space constraints. I tried the same setting as award number and the display didn't change. (Award number is already being wrapped even with that minimum width setting. )

PR #823, opened 2018-11-16
Fix workflow for web-linked only repos

karenhanson:
This fixes an issue where if a preparer prepares a proxy submission that contains a web linked repository only, the submitter cannot submit it.  For this change, I did some changes to the message that appears to confirm which repositories you are submitting to. It now has some wording for web-linked, please let me know if you want me to change the text. 

Fixed a few minor style issues.

Ideally testing would involve looking at the behavior of hitting "submit" on a proxy submission under various circumstances and making sure it behaves as expected. For example, having the preparer prepare the following and see what happens when submitter submits:
1. submission with web-link repo only (no agreements or other repositories so no repositories attached)
2. submission with jscholarship + weblink
3. submission with jscholarship + pmc
4. submission with pmc only

### DIFF
--- a/app/controllers/submissions/detail.js
+++ b/app/controllers/submissions/detail.js
@@ -49,7 +49,7 @@ export default Controller.extend({
    * If the submission is submitted, return external-submissions object from metadata.
    * Otherwise generate what it should be from external repositories.
    */
-  externalSubmissionsMetadata: Ember.computed('model.sub.respositories', 'model.sub.submitted', 'model.sub.metadata', function () {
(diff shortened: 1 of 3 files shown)


htpvu:
@karenhanson github says there is conflict and this can't be merged. COuld you resolve the conflict. 

Additionally, the confirmation box's `Submit` button, i think it'd be better labeled as `Confirmed` instead. Could you chnage the label of that button to `Confirm`? 

with the above points resolved, this PR is mergeable.

karenhanson:
@htpvu I have resolved the conflict, rebased against latest master, and changed "Submit" button to say "Confirm".

PR #825, opened 2018-11-16
File validation and fixes

### LINKED ISSUE #789: Add additional validation on File upload step
when a submission goes back and forth between preparer and submitter, potentially having files uploaded in mulitple session,  a submission can end up having more than one manuscript files. 

Updated per slack discussion: to help users with making better decision about what file to upload, we will display file role description along with each of the existing file's name on the file upload page

birkland:
* Adds validation that requires _exactly one manuscript_ to proceed to 'next' from the file upload page
* Allows file details (role, descriptions) of already-upload files to be editable
* Consolidates all files (pending, and existing) into a unified table
* Places the file upload box at the bottom of the page, as it seemed a more logical location with the above consolidation, except in the case of a proxy submitter intentionally preparing a submission without a manuscript
* Fixes pre-existing bugs and logic errors, including:
  * Improper counting of the list of files, which allowed the existing file count check to be bypassed by navigating back and forward
  * Wrong constant for supplemental files, resulting in Deposit services failing to deposit any submissions that contained supplemental files
* Adds a file validation step on final submission, requesting the user to go back and edit the files if the manuscript count is incorrect.

To test:
* As a proxy submitter, try adding and deleting files, setting their role, description, etc.  Verify that the only way to proceed to the "request approval" stage is to have exactly one manuscript, with any number of additional files
* As an approver/submitter, edit the files created by the proxy submitter.  Change their role and description, or delete them.  Verify that those changes persist, and the only way to proceed to the "submit" stage is to have exactly one manuscript, with any number of additional files.
* As a proxy, create a submission that does _not_ have any manuscripts attached.  Verify that a popup asks you to confirm (and warns that the submitter will have to add one), and submit for approval.  As the submitter, try immediately submitting that submission.  It should warn that a manuscript is necessary.  Either request changes, or fix it yourself.  Verified that the fixed submission is accepted.

Resolves #789

### DIFF
--- a/app/components/workflow-files.js
+++ b/app/components/workflow-files.js
@@ -1,8 +1,10 @@
 import WorkflowComponent from './workflow-component';
+import { inject as service } from '@ember/service';
 
 export default WorkflowComponent.extend({
   store: Ember.inject.service('store'),
   files: Ember.A(),
+  currentUser: service('current-user'),
   init() {
     this._super(...arguments);
     this.set('files', Ember.A());
@@ -17,11 +19,7 @@ export default WorkflowComponent.extend({
       let manuscriptFiles = [].concat(this.get('files'), this.get('model.files') && this.get('model.files').toArray())
         .filter(file => file && file.get('fileRole') === 'manuscript');
 
-      if (manuscriptFiles.length == 0) {
-        toastr.warning('At least one manuscript file is required');
-      } else if (manuscriptFiles.length > 1) {
-        toastr.warning(`Only one file may be designated as the manuscript.  Instead, found ${manuscriptFiles.length}`);
-      } else {
+      let updateExistingFiles = () => {
         // Update any *existing* files that have had their details modified
         let files = this.get('model.files');
         if (files) {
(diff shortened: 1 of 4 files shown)


htpvu:
This works exactly as described. it is quite nice. 

I do have one concern about the proxy submission needing one manuscript before he can submit. I'm thinking about the usecase where proxy does not have a manuscript at hand, but create the submission for submitter to finish off by adding the manuscript file prior to submission. Let me stew on this a little bit.

birkland:
How about a model popup if a proxy submitter proceeds without a manuscript "there is no manuscript attached to this submission.  That means the designated submitter will need to provide one.  Are you sure you wish to continue without a manuscript?". 

The problem then is how to force the submitter to add one.  Maybe it's possible to put the submission in a "manuscript-required" state, instead of approval-requested?  The UI already handles manuscript-required.  The caveat here is that the submissionEvent would still be approval-requested, as would the notification message.  That's probably fine, though.  I think the end result for manuscript-required  is that when somebody clicks on the link to see the submission, it goes straight to editing (rather than displaying the option to submit, cancel, or edit)

PR #848, opened 2018-12-19
Adds property to support Ember development in Windows, updates README

karenhanson:
Adds `"watcher":"polling"` property to `.ember-cli` to support reloading in Windows.

### Testing
This should be tested in Windows, but also *nix to ensure it does not break things for other users.
* Follow modified README instructions for starting up `docker`. 
* Once the containers are up, confirm you can load the Ember app and log in
* Change something that you can easily validate through the GUI in the Ember `/app/` folder, observe that the docker ember log reports a change to a file and reloads.
* Hard refresh the page in the browser, you should be able to see your change.

### DIFF
--- a/.ember-cli
+++ b/.ember-cli
@@ -7,5 +7,7 @@
   */
   "disableAnalytics": false,
 
-  "testPort": 4200
+  "testPort": 4200,
+  "watcher": "polling"
+
 }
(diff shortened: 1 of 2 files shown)


emetsger:
Worked for me too!

karenhanson:
Hurray!!

PR #852, opened 2019-01-24
Mock store example [Do not merge]

markpatton:
This pr demonstrates how the store service can be mocked to test a component. This means that a great deal of functionality can actually be tested without the shib environment and all those complications.

One takeaway is that code modules which directly make http requests should be rewritten to use services instead. For example workflow-basics directly makes DOI requests. This makes testing code paths which make a DOI request hard if not impossible. Instead there should be a DOI service which understands how to validate DOIs and retrieve metadata about them. Then tests can inject a mocked DOI service.


htpvu:
outcome of the work done in this ticket https://github.com/OA-PASS/pass-ember/issues/855

htpvu:
used as example in new tests. can be closed now.

PR #861, opened 2019-02-01
Refactored the submission workflow to be on 8 routes instead of 1

### LINKED ISSUE #853: Break up submission workflow into separate routes in preparation for CMF work

karenhanson:
PASS was initially written with the entire submission workflow on one route, `submissions/new`. Having the entire workflow on a single route meant that each tab does not have it's own route/controller layer for loading and processing data specific to the tab.  As the application becomes more complicated and data intensive, the `submissions/new` controller shared by 7 tabs is getting very large, making changes difficult to implement. To address this, the following changes were made. They had to be done together to preserve the functionality of PASS:

#### Route changes
This change is focused on improving the route structure. It splits the single new route into 8:
1.  A parent `submissions/new` route. Minimal changes have been made to the parent route/controller with the emphasis in this first PR on preserving functionality. All submission workflows begin by visiting this route. It initializes the submission, resetting all properties.
2. A route for each tab in the workflow, 7 in total e.g. `submissions/new/basics`. All that the route objects for these do at the moment is set the step number property for the tab display. Each route has a controller associated. All that the controller contains at the moment is instructions for the loadNext/loadPrevious action and computed properties to preserve interactions between the child route and the original new route. Some thought should be given to making these more dynamic in the future to support situations where the tabs might vary for different kinds of submissions - for now they are hardcoded. Eventually, the routes and controllers can be refactored to contain relevant tab-specific model, processing, and validation. For now they are there to proved the framework for future improvements.

#### Workflow service
To simplify this restructuring, a new `workflow` service has been created to hold properties that used to exist in the `workflow-wrapper`. When there was one route the `workflow-wrapper` component loaded the relevant tab component through a series of `{{if}}` statements in its template, and could be used to pass data into all components. With 7 routes, the information does not persist on each tab so it is no longer viable for storing workflow properties. 

The workflow service is a bunch of setters and getters that replace the job that the `workflow-wrapper` used to do. We can either continue with this approach if it seems beneficial, or add properties to the parent controller - this will be addressed (or not if it is viable as is) in a future change. 

#### Component layout
Where possible, I have preserved the component names to make it easier to see the changes in the files. I have stripped out the `WorkflowComponent` logic that automated some of the tab logic. 
As mentioned, the `workflow-wrapper` component used to load the tab UI components using `{{if}}` in the template. Now the `workflow-wrapper` just deals with the visual wrapper (`submission-nav` and comments display). Instead, the default template for each of the new tab routes initiates the template/component that has the display logic. There may be a way to do away with a layer of components, but this was the simplest approach for preserving as much code as possible in this first PR. 

#### Remove grants query from new route model
While refactoring it became clear that the grants query in `submissions/new` route's model wasn't being used. I deleted it to get rid of the code redundancy, besides which, now that the grants tab has it's own route, it could be part of the grant route model.

#### Add isProxySubmission to Submission model
The `submissions/new` route has a property called `hasProxy`.  This should be migrated to a computed property on the Submission model, but due to the way `submitterEmail` and `submitterName` are currently being managed in the workflow it will be implemented on a separate PR. For now, this was added to the model for use by the `workflow-wrapper`, otherwise there would have been a lot of code to pass this into every instance of workflow-wrapper.

#### Broken tests
Note that tests have not been updated to reflect the new model since extensive refactoring is to take place over the coming weeks and will require them to be rewritten. Default tests were created for new routes, controllers, and components.

closes #853


markpatton:
Looks good. Very impressive work. I think this results in some good simplification.

I am not a fan of storing application state in a service and think it should generally be avoided. Instead the parent route (new)  and its controller could be used. But either case would be a huge improvement!

Perhaps some of the hasProxy calculation stuff might belong in the workflow service? It is hard to untangle. I like having it on the submission model as much as possible.

karenhanson:
Thanks!

Indeed, I was not at all sure about the workflow service approach, but I did this to make it easier to take workflow-wrapper out of the picture. I think doiInfo could go on doi service, some things can be moved to the submission model as computed properties, and then it might be easier to get rid of it. 

hasProxy - I plan to move to the Submission model (I actually added it as isProxySubmission) - I needed a few of the properties in `submissions/new` to calculate it so couldn't move it without code disruption, my next PR will be to make that so!

app/components/policy-card.js
```diff
@@ -16,8 +32,14 @@ export default Component.extend({
   }),
   didRender() {
     this._super(...arguments);
-    if (this.get('methodAJournal') && !this.get('removeNIHDeposit')) {
```
emetsger: Is the `removeNIHDeposit` condition no longer applicable after refactoring?

karenhanson: I did a light sweep of removeNIHDeposit property because it was appearing in multiple places and there were two properties doing very similar things - it's one of the few I was forced to refactor a bit and ultimately eliminate. The removal happens on `app/components/workflow-repositories.js#53`.  The logic does the same thing. If pmcPublisherDeposit is true, it will remove the NIH repository. If the journal is methodA, pmcPublisherDeposit should always be true.

karenhanson:
@emetsger I would like to do a bit more research on it now that I have a better vocabulary for ember and have a chance of doing a meaningful search. The workflow service was a convenient band-aid, and I will spend the next week reducing the need for it. Right now it's a mix of stuff, but I want to narrow it to hold UI state relevant for the duration of the workflow but not reflected in the submission model e.g. radio button status, tab state etc. We can see what's left then maybe. 

The problem of parent controller properties going down to child components is simply that you have lots of code purely for passing things up and down the tree... but I may be able to find a less burdensome way to do it. I'd thought about a model for non-persisted data, but have no idea if that's an appropriate use of the Ember model.

In principle it seems right that properties should be passed into the components from outside, making it easier to test, regardless, I will probably move workflow service interaction up the tree and lighten the component.

karenhanson:
Merging this to a refactoring branch

Issue #854, opened 2019-01-25
Test new submission workflow routes


htpvu:
follow up of #853

htpvu:
Karen has done the first iteration of work, is out of office. May be assigned to others.

PR #872, opened 2019-02-23
Add tests for workflow controllers

karenhanson:
This PR:
* Adds unit tests to `submissions\new`, and the controllers for its sub-routes.  
* Adds a missing test file for `workflow-basics-user-search`
* Adds some minor changes to `workflow-basics` to support integration test that will be written
* Removes some stray console logs.

Interesting lesson learned: some controllers use the `swal` popup. This will actually popup in tests if not overridden like this, for example:
`swal = result => new Promise(resolve => (assert.ok(true)));`
If we develop these tests in the future, it would be useful to configure the swal property to return different responses (ok, cancel etc). For now, I am submitting as is so that this work can move forward.

Part of #854


markpatton:
I ran the tests successfully and read through them. Looks good. A fair amount can actually be tested!

PR #906, opened 2019-04-22
Use new doi service

### LINKED ISSUE #899: Update ember code to pass the DOI CrossRef response to the journal service to retrieve journal object for the submission workflow

markpatton:
Updates the doi service resolveDOI method to use the new journal service.

- [x] Update doi service
- [x] Unit test new functionality
- [x] Remove NLMTA service
- [x] Ensure ISSN structure in metadata blob is correct
- [x] Add $schema to metadata blob
- [x] Manually test against new journal service

In order to test, start the docker environment, do a submission, and check that you are able to retrieve metadata from a DOI and that the metadata blob is as expected after submission. In particular look at the issn field.

Closes #899 #911


markpatton:
Consider moving nlmtaservice in server side doi service

PR #914, opened 2019-05-03
Policy service

jabrah:


app/services/policies.js
```diff
@@ -0,0 +1,113 @@
...
+   * from the Ember Store, then add the model object to the reference when done.
+   *
+   * @param {array} refs reference objects {
+   *    url: '',
+   *    selected: true|false
+   * }
+   * @param {string} type object type (example: 'repository')
+   * @returns {array} where Repository ID resolved to an Ember model object
+   */
+  resolveReferences(type, refs) {
```
markpatton: Code which uses resolveReferences could be simplified if it returned a promise that resolves to the array of objects using Promise.all.

PR #934, opened 2019-05-16
Re-enable ability to edit draft submissions

### LINKED ISSUE #905: Re-enable editing of draft submissions
A few things need to happen before this will be done:

* Journal service integrated : this is needed because it will overhaul the handling of DOI data and how various related objects are created/handled in the submission workflow
* Publication object is created and persisted in the `workflow-basics` step. The publication object is currently created in the `submissions/new` route, technically outside of the workflow steps. It is currently not persisted until the submission is finally submitted in the review step.
  * This is supported already by `doi#createPublication()` but needs to be integrated along with the Journal service work

Completing this will enable the Edit button in the submissions table and submission details page. It will also allow draft submissions to display correctly in the submissions table, since there will actually be a title to show.

jabrah:
Closes #905 

No tests yet!

My biggest concern is the way Journal data is loaded in the case of a draft submission with a Publication with no DOI. This is handled similarly to how DOI would be loaded into the widget: if applicable, it will send the `selectJournal` action to kick off it's logic and add the Journal metadata to the workflow service. Applicability here is if a Journal object is passed to the component from the `submission/new` route, which should happen only when loading draft submissions.

### DIFF
--- a/app/components/display-metadata-keys.js
+++ b/app/components/display-metadata-keys.js
@@ -6,7 +6,14 @@ export default Component.extend({
   ignoreList: ['agent_information', 'external-submissions'],
 
   metadata: Ember.computed('submission', function () {
-    return JSON.parse(this.get('submission.metadata'));
+    try {
+      return JSON.parse(this.get('submission.metadata'));
+    } catch (e) {
+      /**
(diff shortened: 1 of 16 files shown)


jabrah:
Closing for now, will re-open from a different branch

PR #937, opened 2019-05-21
More properties should be added to the list of 'valid' fields to take from Crossref

### LINKED ISSUE #936: Include the DOI of the published article in the metadata blob
If a user enters a DOI during submission, it ought to be captured in the metadata blob.  Prior to CMF implementation, DOI was captured, and some agents may expect it to be there.  

This is the case with respect to the NIH package provider: it will include the article DOI in package metadata when preparing and depositing a package to NIHMS.  

For reference, the DOI is documented in the [global](https://github.com/OA-PASS/metadata-schemas/blob/master/schemas/jhu/global.json) schema.

If we opt not to capture the DOI, we should discuss the rationale and document it here.

jabrah:
Closes #936 

A single very quick submission suggested that this fix at least adds DOI to the metadata blob. Unit tests and probably more testing is required to verify.

This takes the first approach outlined in the ticket to automatically try to inspect the given set of schema for properties to add.

* [x] Write tests to capture this behavior

### DIFF
--- a/app/services/metadata-schema.js
+++ b/app/services/metadata-schema.js
@@ -110,9 +110,24 @@ export default Service.extend({
 
     schemas.map(schema => this.alpacafySchema(schema))
       .forEach((schema) => {
-        Object.keys(schema.options.fields)
+        Object.keys(schema.schema.properties)
           .filter(key => !fields.includes(key))
           .forEach(key => fields.push(key));
+      });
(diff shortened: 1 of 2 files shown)


jabrah:
I think it was determined that we include this workaround now, with a more permanent fix later

PR #941, opened 2019-05-29
Reenable draft edit

### LINKED ISSUE #905: Re-enable editing of draft submissions

jabrah:
Closes #905 

------

_Reopening this PR from the `pass-ember` repo, instead of from my fork so that other people can make changes._

_Only limited tests so far._

_My biggest concern is the way Journal data is loaded in the case of a draft submission with a Publication with no DOI. This is handled similarly to how DOI would be loaded into the widget: if applicable, it will send the `selectJournal` action to kick off it's logic and add the Journal metadata to the workflow service. Applicability here is if a Journal object is passed to the component from the `submission/new` route, which should happen only when loading draft submissions._

------

## Changes

* Of course, the "Edit" button for draft submissions is now enabled
* Publication objects now persist after the workflow basics step
  * This fixed a bug where a user was not able to navigate to the submission details page of a draft submission (there was no Article link to click in the submissions table)
* Files are now uploaded immediately to Fedora in the workflow files step
* User entered metadata is maintained through edits
  * Metadata is only saved to the submission after the user leaves the workflow metadata step; the user has to click Next to bring up the workflow files step
* Enhance the "delete draft submission" feature
  * Publication object and all associated File objects are deleted, then the Submission is deleted
* Plenty of code cleanup
* A couple unrelated bugs fixed
  * Livereload script shouldn't be tried when loading the app
  * TransitionAborted errors are ignored. There was a bug where this error appeared every time a user entered the submission workflow, or any time a URL query param is modified

### DIFF
--- a/.ember-cli
+++ b/.ember-cli
@@ -6,7 +6,7 @@
     Setting `disableAnalytics` to true will prevent any data from being sent.
   */
   "disableAnalytics": false,
-
+  "liveReload": false,
   "testPort": 4200,
   "watcher": "polling"
 
--- a/app/components/display-metadata-keys.js
+++ b/app/components/display-metadata-keys.js
@@ -6,7 +6,14 @@ export default Component.extend({
   ignoreList: ['agent_information', 'external-submissions'],
 
   metadata: Ember.computed('submission', function () {
-    return JSON.parse(this.get('submission.metadata'));
+    try {
+      return JSON.parse(this.get('submission.metadata'));
+    } catch (e) {
+      /**
+       * We just need to handle the case where 'submission.metadata' does not exist.
+       */
+      return {};
+    }
   }),
 
   // TODO: could be changed to get the real label from relevant metadata schema!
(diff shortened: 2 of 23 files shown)


markpatton:
I made a few cleanups and fixes. Looks good to me now.

jabrah:
Files are not yet persisted in a Draft submission yet. This means that if a user uploads a file (and clicks Next to get to the Review step), but then leaves the workflow, when they go to edit the submission, the file will no longer be attached to the submission. The user will have to re-add the file(s).

Workflow should be modified to persist the File objects when the submission is auto-saved (currently happens when the user transitions between steps).

PR #950, opened 2019-06-05
Proposed solution to the 403 issues around deleting Submission objects

### LINKED ISSUE #949: dev instance - deleted drafts don't go away
this happens in dev with issue 17 
steps:
1. try to create submission, see error trying policy service
2. look at submission - see an incomplete submission
3. delete this incomplete submission
4. try another submission, get same policy service error
5. look at submission list, see both incomplete submissions

jabrah:
Closes #949 

Instead of issuing `DELETE` requests to Fedora for a submission and associated objects, simply set it's `submissionStatus` to `cancelled`. User experience should be the same as actually deleting the record, however, it means that _all_ draft submissions are kept in Fedora. All Publication objects and potentially uploaded files will also be kept.

We now need to filter out `cancelled` submissions from the submissions table. This is done in the  `/submissions` route model.


emetsger:
I need to refresh my memory about what notifications, if any at all, are emitted when a submission is cancelled.

At any rate, I need to review notification services, insuring that notifications aren't emitted when a draft is cancelled.

emetsger:
After a review and refresh of Notification Services-

NS only uses `SubmissionEvent` for making decisions about issuing notifications.  It doesn't use `Submission.submissionStatus` at all.

In the future, if we decide to issue a `SubmissionEvent` when a submission is "deleted", then there'd need to be a conversation as to which `SubmissionEvent.EventType` is used, and how that would impact NS.

markpatton:
Also consider bumping ember-fedora-adapter to 0.0.5 as part of this.

jabrah:
We must be careful about dependencies in `package.json`. It looks like `master` has a version of jQuery that may break alpaca

PR #956, opened 2019-06-11
First draft at polling ES for changes

### LINKED ISSUE #955: Submissions killed with the "Abort" button in the workflow still appear in the submissions table (until a page refresh)
This happens because of asynchronous behaviors in our stack.

* User clicks the Abort button to kill the draft submission they are currently working on
* `submission-handler` sets the draft submission's status to `cancelled` (no SubmissionEvent is created)
* Ember app then transitions to the `submissions` route

At this point the dead draft appears in the table showing all submissions, and it appears as a draft that can be edited (or deleted). This ultimately occurs because the `submissions` route is populated by Elasticsearch query. Since the application transitions as soon as the Submission is marked as `cancelled` in Fedora, Elasticsearch does not have enough time to perform its indexing. The ES query used to populate the submissions table will see the old submission in ES, not the updated submission.

jabrah:
Closes #955 

A shared service is created that can be used to maintain an "ignore list" that can be referenced before and after a route transition. This is sometimes necessary because of the lag time of reindexing in Elasticsearch.

This work focuses on "Aborting" draft submissions from the workflow. The draft submission is added to the ignore list, then this list is seen by the `submissions` route and added to the ES query to prevent the given ID from appearing in the ES results.

### DIFF
--- a/app/controllers/submissions/new.js
+++ b/app/controllers/submissions/new.js
@@ -71,6 +71,7 @@ export default Controller.extend({
     },
     abort() {
       const submission = this.get('model.newSubmission');
+      const ignoreList = this.get('searchHelper');
 
       swal({
         title: 'Discard Draft',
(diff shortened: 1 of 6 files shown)


jabrah:
Will switch to the other approach of maintaining an ignore list in a shared service

PR #962, opened 2019-06-18
Set metadata provided to metadata step as readonly in forms

### LINKED ISSUE #939: Make sure that if fields in the metadata form is populated by Cross ref look up, they are not editable in anyway
This applies to all the metadata forms.

### LINKED ISSUE #965: NLMTA ID doesn't make it into the submission metadata
`journal-NLMTA-ID` should come from the Journal object

jabrah:
Closes #939 
Also Closes #965 

One caveat is provided in the ticket. Alpaca warns about invalid data for _Authors_ and _ISSNs_. This doesn't seem to effect the metadata that gets saved to the submission.

`workflow-metadata` now assigns all keys that appear in the incoming metadata as read-only. When these keys are encountered in any rendered schema, they are marked as read-only in Alpaca. This implementation seems to work for submission with and without DOIs.

## Some test cases
This example test case should exercise basically everything, so is extensive:

1. Start new submission
2. Enter a DOI, example: `10.1039/c7an01256j`
    * There should be a (hopefully) brief popup in the top right saying that we're looking up your DOI
    * The Next button should be greyed out while the DOI is loading
    * After the DOI loads, the Title and Journal fields should be filled in and read-only
3. Click Next, add some grants
4. Proceed through the workflow until you get to Step 5. Details (the metadata forms)
    * There should be plenty of data filled into the metadata forms now. We should verify that:
      - Any value that has been prepopulated from the DOI is read only
      - If there are any authors and/or ISSNs filled in, the user is not able to add or remove any more values in these fields
5. Click Next until you get to the Files step (the metadata will then be saved to the draft submission)
6. Leave the workflow by clicking the Submissions tab in at the top of the page
    * The submission you were working on should now appear at or near the top of the submissions table, be marked as _draft_ and have _Edit_ and _Delete_ options
7. Click Edit to edit the draft submission you just made. You should be taken to Step 1 of the workflow
    * Your DOI should be shown, and Title and Journal should be displayed as read only
    * There should briefly be a popup saying that we are reading your DOI, and the Next button should be disabled
    * Once the DOI service finishes, a green popup should be shown and the Next button will be enabled
8. Click Next until you get to the Details (metadata) step
    * Verify that data is present here as expected - it should look like it did last time you were in the workflow
    * You don't need to keep going, since there are no changes to the metadata
9. Go back to the first step of the workflow, either by clicking Back until you get there, or you can click **1. Basics** tab in the navbar
10. Delete the DOI
    * Title and Journal should be cleared
11. Enter a value for Title, and pick a Journal
12. Proceed through the workflow until you get to the metadata step
    * Article Title and Journal/ISSN info should be filled out, these fields should be editable

### DIFF
--- a/app/components/workflow-basics.js
+++ b/app/components/workflow-basics.js
@@ -97,9 +97,23 @@ export default Component.extend({
     if (workflow.isDataFromCrossref()) {
       this.get('workflow').setFromCrossref(false);
       this.set('doiInfo', {});
-      // debugger
-      console.log('Reset Crossref stuff');
+      this.clearPublication();
     }
+  },
+
+  /**
+   * Remove all data from the current Publication
+   */
+  clearPublication() {
+    this.get('publication').setProperties({
+      doi: '',
+      title: '',
+      abstract: '',
+      volume: '',
+      issue: '',
+      pmid: '',
+      journal: undefined
+    });
   },
 
   actions: {
--- a/app/components/workflow-basics.js
+++ b/app/components/workflow-basics.js
@@ -45,30 +45,6 @@ export default Component.extend({
     const hasPublication = !!(publication && publication.get('doi'));
 
     this.lookupDoiAndJournal(!hasPublication);
-    /**
-     * See IMPL NOTE above regarding the existance of submission metadata
-     */
-    // try {
-    //   const md = JSON.parse(this.get('submission.metadata'));
-
-    //   /**
-    //    * Set workflow doiInfo because of some weird timing between `routes/submissions/new/index#beforeModel`
-    //    * and `routes/submissions/new#model()` causing the doiInfo in 'workflow' to reset after it is
-    //    * defined by the incoming submission
-    //    */
-    //   this.get('workflow').setDoiInfo(md);
-
-    //   /**
-    //    * Check to see if incoming metadata is an empty object. If metadata is an empty object, then
-    //    * we still need to load DOI, but may not have to set the returned Publication object.
-    //    */
-    //   if (Object.entries(md).length === 0 && md.constructor === Object) {
-    //     this.lookupDoiAndJournal(!hasPublication);
-    //   }
-    // } catch (e) {
-    //   // Either 'metadata' is missing or malformed
-    //   this.lookupDoiAndJournal(!hasPublication);
-    // }
   },
 
   /**
--- a/app/components/workflow-basics.js
+++ b/app/components/workflow-basics.js
@@ -2,23 +2,12 @@ import Component from '@ember/component';
 import { inject as service } from '@ember/service';
 
 /**
- * IMPL NOTE:
- * An interesting thing to consider: say a user goes through the submission workflow and ends up filling
- * out some custom values in the metadata fields presented to them. The user then leaves the workflow
- * leaving us with a 'draft' submission that can be resumed later. When we re-enter the submission
- * workflow, we don't want to clobber the user-entered metadata.
- *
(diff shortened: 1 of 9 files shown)


jabrah:
There is a downside of this simplistic approach. When editing draft submissions or even "finishing" submissions, this approach will make all previously entered metadata read-only. I don't think this is desirable behavior

htpvu:
Yeah making anything that didn't come from CrossRef as Read ONly  is not desirable. Also  sounds like Alpaca was giving warnings? also not desirable. 

is there a complicated approach?

jabrah:
To clarify, the warning only appear in the JS console, not something that is noticeable be most users.

The alternative is to identify when metadata comes from Crossref and only then set fields to read only maybe just key off the presence of a DOI?

jabrah:
New changes modify behavior in not-insignificant ways

* `workflow-basics` : the first step of the submission workflow
  * `submission.metadata` used to be cleared on this step for some reason
  * If a DOI is present on the Publication object, this step will now _always_ make a call to the DOI service. The results are now separate from the original submission's metadata and are put into `workflow.doiInfo` so it can be maintained through the workflow
* `workflow-metadata` : metadata (details) step
  * Can now have a _read-only_ list that holds a list of keys that should be marked as read-only in any rendered Alpaca form
  * The _read-only_ list is pulled from `workflow.doiInfo` iff it is marked as coming from Crossref `workflow.isDataFromCrossref`
* `doi` : the client part of the DOI service
  * Made to be more defensive about copying certain properties. Because of the way metadata is merged, values of _'undefined'_ were being put into the metadata blob, overwriting already existing data. 
    - Data is only removed while merging blobs if _keys_ don't exist. If a key exists, but has an undefined value, then 'undefined' is merged into the current metadata blob

markpatton:
I am able to remove read-only entries with multiple values like ISSN.

markpatton:
When I entered a doi, went through the workflow to the end, then went back to the start and deleted the doi, in subsequent steps the data was still marked read only.

jabrah:
More changes added to address issues raised by @markpatton . I ended up simply clearing the submission metadata when the DOI is removed to avoid potential metadata conflicts. There is also a change to the way the JSON schema validator is configured that decreases performance, a note was left along side the config change in code.

jabrah:
The way `journal-NLMTA-ID` is mapped is also changed because I was already modifying that code. The mapping should now check the Journal object for it's nlmta data and add it to the metadata blob if found.

jabrah:
http://www.alpacajs.org/docs/fields/array.html

suggests that we can set `options.toolbarSticky: false` when we want to hide the toolbars that appear when hovering over array rows

jabrah:
Another issue crops up with this PR:

This solution relies on a response from the DOI service, meaning that if a user clicks through the Step 1 form before the DOI service can respond, it will not behave as expected. For example, a user enters a DOI and data is filled out as expected (DOI service responds and the application does it's thing). The user then proceeds to go through the workflow and sees a bunch of read-only data in the metadata forms, again as expected. Say the user then leaves the workflow with this draft autosaving.

The user starts editing the draft submission. Step 1 opens and the user sees the DOI, title, and journal filled out. With the PR in it's current state, some trouble may begin here. If the user clicks Next too quickly, the flag needed to mark the DOI data as read only is not set, because the DOI service will not have yet returned. 

Basically, I think we need to hold the user until the DOI service returns. This will give us a chance to address potential DOI service errors, too. Crossref seems a bit finicky. At least during my testing, it seems to be very slow sometimes, possibly an issue with Crossref throttling requests.

PR #975, opened 2019-06-26
Grab publication date from DOI data

### LINKED ISSUE #964: `publicationDate` does not make it into the metadata blob
There is a "Publication Date" field in our metadata schema, but no such date makes it from the Crossref blob we get from the DOI into the metadata forms - thus this field doesn't make it into the submission's metadata blob.

`doi#doiToMetadata` has the mapping which needs to be updated once we know which Crossref field matches the `publicationDate`

jabrah:
Closes #964 

This implementation relies on an updated schema service, as it changes `publicationDate` to a string, from a date format.

* Map crossref `issued['date-part']` field to `publicationDate`
* Temporarily bump schema service version, will be updated once the new version is released

This is actually in a state that can be tested, simply pull these changes, bring up the `pass-ember` docker environment, start a new submission with a DOI of your choice, proceed through the submission workflow. Publication date should now appear in the metadata forms, and it should not trigger any validation errors.

### DIFF
--- a/.docker/docker-compose.yml
+++ b/.docker/docker-compose.yml
@@ -31,7 +31,7 @@ services:
       - activemq
 
   schemaservice:
-    image: oapass/schema-service:v0.4.0@sha256:f6662809566184c9332d394465bea2cf93080aa018cf8ef322d0d6cb99982f17
+    image: oapass/schema-service:v0.4.0-1-g5227dd7@sha256:aff04c21f70d13bb9a9113e9b83b93b2a6e0b52c6d01fd322769379359c96c60
     container_name: schemaservice
     env_file: .env
     ports:
--- a/app/components/workflow-metadata.js
+++ b/app/components/workflow-metadata.js
@@ -114,7 +114,7 @@ export default Component.extend({
   actions: {
     nextForm(metadata) {
       const step = this.get('currentFormStep');
-      this.updateMetadata(metadata);
+      this.updateMetadata(metadata, true);
 
       const schemaService = this.get('schemaService');
       const schema = this.get('schemas')[this.get('currentFormStep')];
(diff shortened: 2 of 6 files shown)


emetsger:
This is a metadata blob I received, and it seems to use a comma as a separator between the date elements.  Perhaps because `date-parts` in the DOI service return is also comma separated?

- DOI: `10.1186/1465-9921-9-59`
- publicationDate: `2008,8,7`

```json
{
  "publisher": "Springer Nature",
  "issue": "1",
  "title": "Signaling pathways required for macrophage scavenger receptor-mediated phagocytosis: analysis by scanning cytometry",
  "volume": "9",
  "journal-title": "Respiratory Research",
  "issns": [
    {
      "issn": "1465-9921",
      "pubType": "Print"
    },
    {
      "issn": "1465-993X",
      "pubType": "Online"
    }
  ],
  "authors": [
    {
      "author": "Timothy H Sulahian"
    },
    {
      "author": "Amy Imrich"
    },
    {
      "author": "Glen DeLoid"
    },
    {
      "author": "Aaron R Winkler"
    },
    {
      "author": "Lester Kobzik"
    }
  ],
  "journal-NLMTA-ID": "Respir Res",
  "publicationDate": "2008,8,7",
  "doi": "10.1186/1465-9921-9-59",
  "$schema": "https://oa-pass.github.io/metadata-schemas/jhu/global.json",
  "agent_information": {
    "name": "Chrome",
    "version": "74"
  }
}
```

jabrah:
Turns out that `date-part` is a nested array for some reason

jabrah:
Further changes:

* Update the way metadata blobs are merged together, allowing for fields to be deleted during a merge (which was not the case before, meaning users couldn't remove a metadata field once it was saved...)
* Flattened the Crossref date data for `publicationDate` to avoid weird formatting in the blob

PR #983, opened 2019-07-03
Try adding and removing repos when necessary

### LINKED ISSUE #982: Effective policies isn't updated correctly in some situations
* Start a submission, add an NIH grant and a non-NIH grant
* Go through workflow until you get to the metadata forms
* Go back to Grants step and remove the NIH grant
* Go through the workflow
  * At the Repositories step, the UI would error out and reload

This bug occurred because the UI was not updating `submission.effectivePolicies` when grants were selected or removed. Instead only the policy cards themselves were able to add the remove themselves. This was done initially to allow users to remove the NIH policy by selecting the option saying that the publisher had/has taken care of the submission. However, the policy step would not create an NIH card that could remove the policy from `effectivePolicies` because the NIH grant was no longer associated with the submission. Since there was now an erroneous "effective policy" that could no longer be removed, the policy service saw this and threw an error, as it should.

There is an a somewhat error in the Repositories step where current repositories were not being updated if repositories were already associated with the submission. This meant that if a user had gone through the workflow and set one or more repositories, but then later went back and added or removed a grant (associated with a different repository), then the action would not have any consequences on the repositories associated with the submission.

The two bugs described above are both related to the fact that changes to grants in an in-progress submission isn't currently propagated properly through the submission object.

_PS sorry about how verbose this is getting - I just need to organize my thoughts here_

jabrah:
Addresses part of #982 

This PR has multiple changes centered around modifying effective policies and repositories on an active submission in a workflow. Description of the buggy behavior can be seen in the linked ticket. A few changes were made to get around the issue:

* The Policies step will now fully refresh `submission.effectivePolicies` every time the Policy step is run based solely on the results of the policy service
  * The route clears effective policies before invoking the policy service. Any policy will appropriately be added back to the submission during this step. This removal and (re)adding of policies is done entirely client side in the Ember store, then persisted when the user clicks the Next button.
* The Repositories step is now changed to compare `submission.repositories` to the repo list from the policy service, taking the policy service results as truth.
  * Repositories on the submission that do not appear in the results will be removed from the submission
  * Repositories that are not present on the submission, but appear as `required` will be automatically added to the submission.

There is a downside to the approach taken here: if for some reason a user removes a grant from the submission, then clicks Next to save the draft and leaves the workflow, the draft submission will be in a sort of limbo state, where `effectivePolicies` and `repositories` would not have been updated. However, current implementation of editing draft submissions forces the user to step through the entire workflow, so those inaccuracies would be taken care of as soon as they make it to the appropriate step.

### DIFF
--- a/app/components/workflow-grants.js
+++ b/app/components/workflow-grants.js
@@ -1,8 +1,20 @@
 import Component from '@ember/component';
-import { Promise, } from 'rsvp';
 import { inject as service } from '@ember/service';
 import Bootstrap4Theme from 'ember-models-table/themes/bootstrap4';
 
+/**
+ * This component is responsible for displaying a table of grants that are relevant to
+ * the current submission. The list of grants is loaded here and is determined by
+ * the "submitter" of the current submission.
+ *
+ * This step entirely determines the policies that may be applied to the submission,
+ * which appear in a subsequent workflow step. Because of this, any change to selected
+ * grants necessarily must change policies applied to the submission. Adding a grant
+ * MUST trigger a recalculation of applied policies just as much as removing grants.
+ * We can implement this by simply clearing `submission.effectivePolicies` every time we
+ * make a change to grants in this stage. **We could also naively just clear
+ * `effectivePolicies` every time we load into the Policies step...**
+ */
 export default Component.extend({
   store: service('store'),
   workflow: service('workflow'),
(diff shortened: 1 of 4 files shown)


jabrah:
Going to redo this to make it a bit more robust

PR #987, opened 2019-07-08
Modify error handling for DOI service

### LINKED ISSUE #986: Extend DOI validation to include errors from DOI service
Currently the DOI validation only tries simple REGEX check, which will potentially allow through valid-looking DOIs that don't actually point to anything. We should append DOI validation to include potential information from the DOI service, such as the status that informs the client that no such DOI was found in Crossref.

Somewhat related to #922

jabrah:
Closes #986 

Errors are now displayed more plainly to the user when bad things are returned by the DOI service (bad things such as service error messages, or Crossref errors). Error messages now displayed under the DOI input field, instead of a Toastr popup in the top corner

### DIFF
--- a/app/components/workflow-basics.js
+++ b/app/components/workflow-basics.js
@@ -9,6 +9,7 @@ export default Component.extend({
   metadataService: service('metadata-schema'),
 
   inFlight: false,
+  doiServiceError: false,
   isShowingUserSearchModal: false,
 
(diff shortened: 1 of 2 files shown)


markpatton:
Error handling seems to work well. One issue I note is that the doi lookup occurs again and again even when the doi doesn't change. As in click on the box. Not sure if this was the behavior before.

PR #988, opened 2019-07-08
Add spacing between Alpaca radio input and label text

jabrah:
Previous CSS rule actually didn't effect spacing between radio input and label


markpatton:
Looks good.

PR #989, opened 2019-07-11
Address inconsistencies with adding grants

### LINKED ISSUE #982: Effective policies isn't updated correctly in some situations

jabrah:
Closes #982 

This issue ballooned out a bit, addressing not only `submission.effectivePolicies` not being updated, but also `submission.repositories` not being updated properly. As described in the ticket, this occurred when the user goes through the workflow, then goes back to change the grant selection. It also shows up when editing a draft submission and the user changes the grant selection.

## Changes

* workflow service: Can now be used to track grants that are added to the submission. Since data here only survives through one pass through the submission workflow, this information can be compared to the grants list on the submission to determine whether a grant was recently added.
* grants step: the Grants step now adds the grant to the workflow service, as well as the submission object. This way, subsequent steps can determine if a grant was just added, or if it existed on the submission already. This effects the selection status of repositories, for example.
* policies step: effective policies are now simply recalculated every time this step is completed
* repositories step: various factors are used to determine the selection status of `optional` and `choice` repositories
  * Repositories can now be added or removed from the submission, which was not the case before

### DIFF
--- a/app/components/workflow-grants.js
+++ b/app/components/workflow-grants.js
@@ -18,6 +18,7 @@ import Bootstrap4Theme from 'ember-models-table/themes/bootstrap4';
 export default Component.extend({
   store: service('store'),
   workflow: service('workflow'),
+
   pageNumber: 1,
   pageCount: 0,
   pageSize: 10,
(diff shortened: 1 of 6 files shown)


markpatton:
Tests pass. I am able to do a submission and the effective policies and repositories seem to be updated correctly. I was not able to trigger the error described in #982

Issue #1035, opened 2019-08-28
Crossref service down error message for DOI service needs a better display


jrmartino:
If the Crossref service is down, the error mesage which the pass doi servicee passes back to ember is displayed under the DOI entry area, in red, with raw html content - this pobably should be handled in some kind of popup which understands html

jabrah:
A fix has been merged that _should_ render the HTML properly.

PR #1036, opened 2019-08-28
Fix some DOI related user messages

### LINKED ISSUE #1034: Text while waiting for DOI service to respomd needs editing
The popup infobox which informs the user that the DOI service is waiting for a response from crossref is 

 "Please wait while we information about your DOI"

we probably want a verb of some kind before "information"

jabrah:
Closes #1034, #1035 

* No longer escape HTML in error message
* Fix grammar mistake in pending DOI message

To test, first make sure you don't have another Docker stack running. Pull these changes, navigate to the `.docker/` directory, then run `docker-compose up -d` and optionally run `docker-compose logs -f` to monitor Docker. Start a new submission and try messing with the DOI input field to see what happens.

### DIFF
--- a/app/components/workflow-basics.js
+++ b/app/components/workflow-basics.js
@@ -209,7 +209,7 @@ export default Component.extend({
       publication.set('doi', doi);
       this.set('inFlight', true);
 
-      toastr.info('Please wait while we lookup information about your DOI', '', {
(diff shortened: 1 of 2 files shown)


jabrah:
`lookup` should have a space in this case: `look up`

Issue #1049, opened 2020-03-02
Update Ember version


jabrah:
Update version of Ember framework and all related dependencies, such as `ember-cli`, `ember-data`, etc. For efficiency, it is recommended to use [`ember-cli-update`](https://github.com/ember-cli/ember-cli-update)

jabrah:
I did some investigation with the `ember-cli-update` tool. I simply ran all of the codemods available for the current ember version (2.18) and it mostly went well. The only thing to note is that the `qunit-dom` codemod broke a handful of tests. In all of those cases, it changed jQuery element selectors to using `assert.dom('*')`. However, this would consistently fail to pick up any elements. If you remove the `'*'` from the selector, it seems to work as it did before. I assume `assert.dom()` with no argument simply selects the entire document?

I think it would be worth running these codemods before updating to a different version to normalize code before potentially making more drastic changes.

PR #1053, opened 2020-03-09
Upgrade ember to 3

jaredgalanis:
Closes: #1049


app/components/workflow-grants.js
```diff
@@ -201,7 +204,7 @@ export default Component.extend({
         this.get('store').findRecord('grant', event.target.value).then((g) => {
           g.get('primaryFunder.policy'); // Make sure policy is loaded in memory
           this.addGrant(g);
-          Ember.$('select')[0].selectedIndex = 0;
+          $('select')[0].selectedIndex = 0;
```
jabrah: Does not need to be addressed now, but we should note this to be changed to not use jQuery eventually

package.json
```diff
@@ -16,59 +16,63 @@
...
     "merge": ">=1.2.1",
-    "qunit-dom": "^0.6.3",
+    "qunit-dom": "^0.9.2",
     "toastr": "^2.1.4"
   },
   "engines": {
-    "node": "^4.5 || 6.* || >= 7.*"
+    "node": "6.* || 8.* || >= 10.*"
   },
   "dependencies": {
```
jabrah: This is fine for this round. Eventually we should move as much as we can to `devDependencies` instead of `dependencies`

MODULE_REPORT.md
```diff
@@ -0,0 +1,28 @@
+## Module Report
```
jabrah: Do we need to address anything here?

jabrah:
Another bug:

1) Start a new submission
2) At the top, select that you are creating the submission "on behalf of Someone else"
3) Search for "Katrina" and select Katrina Debnam

Once you click on the person's name, you will get an error:

```
Handling error: "Cannot read property 'nextSibling' of null"
TypeError: Cannot read property 'nextSibling' of null
    at clear (runtime.js:126)
    at detachChildren (runtime.js:170)
    at UpdatableBlockImpl.reset (runtime.js:617)
    at Function.resume (runtime.js:224)
    at TryOpcode.handleException (runtime.js:5280)
    at UpdatingVMFrame.handleException (runtime.js:5459)
    at UpdatingVM._throw [as throw] (runtime.js:5178)
    at Assert.evaluate (runtime.js:3049)
    at UpdatingVM.execute (runtime.js:5165)
    at RenderResultImpl.rerender (runtime.js:5489)
```

This takes place in the modal-dialog at <https://github.com/OA-PASS/pass-ember/blob/master/app/templates/components/workflow-basics.hbs#L105-L115>, which kicks off this action <https://github.com/OA-PASS/pass-ember/blob/master/app/components/workflow-basics.js#L141-L145>

PR #1057, opened 2020-03-24
Angle brackets template conversion

jaredgalanis:
- this converts component invocations to angle bracket syntax in templates and tests
- this also works to ensure no implicit `this`'s are used
### TODO:
- [x] run the no implicit this code mod

### DIFF
--- a/.docker/ember-dev/Dockerfile
+++ b/.docker/ember-dev/Dockerfile
@@ -4,6 +4,7 @@ WORKDIR /app
 
 RUN apk add --no-cache nodejs-npm && \
     apk add --no-cache --virtual .build-deps git && \
-    npm install -g ember-cli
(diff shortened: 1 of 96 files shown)


.docker/ember-dev/Dockerfile
```diff
@@ -4,6 +4,6 @@ WORKDIR /app
 
 RUN apk add --no-cache nodejs-npm && \
     apk add --no-cache --virtual .build-deps git && \
-    npm install -g ember-cli
+    npm install -g ember-cli \
```
jabrah: Should make sure there is no trailing backslash, as it kills the Docker container build.
Of course, this image doesn't execute correctly anyway, so we don't really need to worry about this yet

PR #1058, opened 2020-04-16
Fix docker w yarn

jaredgalanis:
- node >= 10 is required for the ember upgrades
- this updates the Dockerfile with the method of installing node / npm as is done in the official node docker image https://github.com/nodejs/docker-node/blob/d071b895dd1b8da6c566f45b971825fe5b087b21/10/alpine3.9/Dockerfile


jaredgalanis:
> I'm amused by the hoops you have to jump through to get yarn to work in Docker. It seems to work fine, though.

@jabrah Yarn was easy enough to get installed in a briefer Dockerfile, what we needed is node 10. For whatever reason such an image was harder to come by.  I know it's a bit verbose, but I figured it is from the official node docker image so I figured it's probably okay.

PR #1076, opened 2020-05-09
adds cancellation refinement to download manuscripts component

jaredgalanis:
- this removes the use of sweet alert to provide loading state
- it moves loading state inline into the file list
- handles cancellation of upload inline as well
- this makes a few minor tweaks to support multiple files
TODO:
- [ ] add new download-service image once https://github.com/OA-PASS/pass-download-service/pull/3 is merged (here and in pass-docker)


jabrah:
Updated `download-service` image will be added in a separate PR

PR #1080, opened 2020-05-12
UPW bugfixes

### LINKED ISSUE #1078: prompts on file upload steps should properly reflect the situation when OA Copies aren't found
When there is no OA copies found, the prompt doesn't read right. We need to either change that prompt to not mention any OA copies, if no OA copies are found OR we need to add some text about OA copies not being found

Screen Shot 2020-05-12 at 10 33 53 AM

jabrah:
Fixes #1078 

* Add message to user when no OA manuscripts are found
  * Previous behavior was to show nothing, leading to inconsistent messaging
* Display full file URL, instead of file name to disambiguate files with the same name

### Testing

* Use DOI: `10.1039/c7an01256j` to check the _Files_ step when with a DOI that has no OA manuscript files
* Use DOI: `10.1038/nature12373` to check behavior with a DOI that has more than one OA manuscript file

### DIFF
--- a/app/components/found-manuscripts/template.hbs
+++ b/app/components/found-manuscripts/template.hbs
@@ -1,81 +1,86 @@
 {{#if this.setupManuscripts.isIdle}}
- {{#if (and (not @disabled) (gt this.foundManuscriptsToDisplay.length 0))}}
-    <div class="border p-4" data-test-foundmss-component>
-      {{#if @disabled}}<h1>Disabled</h1>{{/if}}
+  {{#unless @disabled}}
+    {{#if (eq this.foundManuscriptsToDisplay.length 0)}}
       <p class="text-muted">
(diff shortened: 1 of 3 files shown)


jabrah:
I find the diff a bit confusing for the `found-manuscripts` component. The changes are getting confused about changes to the template conditionals.

In short

```
{{#if (and (not @disabled) (gt this.foundManuscriptsToDisplay.length 0))}}
{{/if}}
```

was changed to 

```
{{#if (and (not @disabled) (gt this.foundManuscriptsToDisplay.length 0))}}
{{#unless @disabled}}
  {{#if (eq this.foundManuscriptsToDisplay.length 0)}}
    <message-to-user-no-manuscripts-found>
  {{else}}
    <previous-component-code-here>
  {{/if}}
{{/unless}}
```

PR #1083, opened 2020-05-22
Fix bug causing UI to be unusable if review popup was cancelled

jabrah:
This is a Draft PR because there are no automated tests yet to check for this issue.

The issue seems to be caused by a lack of handling of cancellations of the SweetAlert popups in the Review step.

When a user clicks the Submit button, an element is made visible that blocks all user interaction with the UI. In "normal" submissions, this is handled after the submission is completed. However, if a user cancels out of the agreement popup, or doesn't click the Agree checkbox, the popups will disappear as expected, but this blocking element will never be removed. The user now cannot interact with the application.

Please feel free to take ownership of this PR and mark as ready for review when it's good to go.

### DIFF
--- a/app/components/workflow-review.js
+++ b/app/components/workflow-review.js
@@ -150,9 +150,9 @@ export default Component.extend({
               return false;
             }));
 
-            $('.block-user-input').css('display', 'block');
-
             this.sendAction('submit');
+          } else {
+            $('.block-user-input').css('display', 'none');
           }
         } else {
           // there were repositories, but the user didn't sign any of the agreements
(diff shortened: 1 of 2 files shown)


jabrah:
A bit more, the number of lines I added might be overkill. I tried to throw in code to explicitly hide the blocking element anywhere that a user could cancel or close a popup. I think I got everywhere, but it is worth checking, since the Sweetalert logic is a bit hairy here.

One more note, putting just a single line to hide the blocker here (https://github.com/OA-PASS/pass-ember/blob/master/app/components/workflow-review.js#L178) will technically work, but I believe is undesirable. IMO we still want the blocking element while the submission is being finalized and persisted, until the transition to the Thank You page.

PR #1122, opened 2022-04-14
Refactoring static assets / pages

jabrah:
Related: https://github.com/eclipse-pass/main/issues/169

* Copy various static assets to Ember app
* Disable static-html in dev docker setup 
* Don't render any references to static pages, such as contact or FAQ pages, that appear in various places in the app
* Update tests to reflect updated config.json

The updated environment variables would have to be ported over to `pass-docker` after this PR is merged, but before a new `pass-ui` image can be generated. 

### Remaining issues
* There are still some references specifically to Johns Hopkins University, especially the institutional OA policy. Not in scope for this ticket to change this
* Need to update the top level proxy to not redirect traffic to `static-html` once that is removed from `pass-docker`. That'll likely be done after this PR is merged, once a new `pass-ui` image is ready to be created. That new proxy image would then have to be brought back into the `pass-ui` local dev environment.

### Manual Testing
This assumes you have an entry in your system's `hosts` file that points `pass.local` to `localhost`
* Boot up the docker env in `.docker/`. 
* The application should be available at `pass.local/app`
* The site should look the same, without nav links to contact, about, or faq pages.
* Make sure there is no notification on screen about failing to load the theme


markpatton:
Looks good. One issue is that env variables will be read at build time by ember, not run time.

app/components/found-manuscripts/index.hbs
```diff
@@ -8,8 +8,11 @@
       <div class="border p-4" data-test-foundmss-component>
         <p class="text-muted">
           We found the following OA copies of your manuscript/article. Please select only a peer-reviewed copy
-          for inclusion in this submission. If you need help determining the appropriate copy to submit to PASS,
-          please <a href="{{assetsUri}}contact.html" target="_blank">contact JHU's Office of Scholarly Communications</a>.
+          for inclusion in this submission.&nbsp;
+          {{#if this.contactUrl}}
+            If you need help determining the appropriate ocpy to submit to PASS, please&nbsp;
```
jaredgalanis, line 13:
```suggestion
            If you need help determining the appropriate copy to submit to PASS, please&nbsp;
```

jaredgalanis (approved):
One small comment, otherwise :shipit:

PR #1127, opened 2022-05-20
 Apply prettier rules to codebase

jaredgalanis:
- this applies the prettier config to the entire code base
- this also sets the printWidth option in prettier config to 120 chars
which matches the checkstyle setting in duraspace
- this fixes on test where line breaking was occuring due to some raw
text in html


jabrah (approved):
Lots of changes! Looks fine from what I checked
Wonder if we should include a recommendation in the README or somewhere to use the `prettier` plugin for your editor of choice

PR #1128, opened 2022-05-23
Add lint to the future and file-based ignores

jaredgalanis:
This addresses: https://github.com/eclipse-pass/main/issues/228

- this adds lint to the future and the es lint and ember template
plugins (https://github.com/mansona/lint-to-the-future)
- this also runs lint to the future ignore which adds ignores to files
so that we can go through and fix them a chunk at a time
- this is advisable so that we can make changes more incrementally

Screen Shot 2022-05-23 at 10 00 39 AM


jaredgalanis:
closing in favor of https://github.com/eclipse-pass/pass-ui/pull/1129

PR #1131, opened 2022-06-02
Upgrade fwd deps

jaredgalanis:
What's in this PR:
- this updates the lock file to include the upgraded version of husky (was left out of a prior PR)
- this fixes the Github action so that ember tests run in CI
- this upgrades the only forwards compat dep in the repo

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -5,9 +5,9 @@ name: Node.js CI
 
 on:
   push:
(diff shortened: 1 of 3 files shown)


jabrah (approved):
I had to clean my local `node_modules` directory (for this project) in order to get tests to pass, maybe because I had been using Node v12 previously? Tests did pass, though, after redownloading dependencies.

PR #1138, opened 2022-09-12
Using enums

jabrah:
* New 'enum'-like attr type for select attributes
* Custom 'enum' transform
* Update appropriate models to use the new type

Notes:
* `store.query` filters still must use the UPPER_SNAKE_CASE values, as they are currently passed directly to the backend
* Enum exports from model files are maintained for reference. We can remove these if necessary


jaredgalanis (approved):
:shipit:

PR #1144, opened 2022-10-10
Fix stale session hang bug

jaredgalanis:
- the session service wasn't being injected which meant `session.invalidate` was never being called
- this caused two bugs:
1) stale sessions would hang on network interactions after becoming stale and the error handler service would trap things in the modal;
2) a re-authentication after this hang would further hang when hitting the `auth/callback` route (probably because the client side session was still in place.
- injecting the session allows the session to be invalidated properly


jabrah (approved):
nice and simple adjustment. Wonder if there's a simple way of testing the invalid/stale session handling?

PR #1151, opened 2022-11-02
Fix som JSON API filters

jabrah:
Resolves https://github.com/eclipse-pass/main/issues/372

A few JSON:API filters needed to be grouped correctly with parenthesis. Some JSON:API filter parts needed to be surrounded by double quotes, if a space could appear in the filter string.

* Submission workflow, Basics step, user search
* Grants list page
* Submissions list page
* Journal autocomplete search

Submissions list page also was not matching against `preparers`


jaredgalanis (approved):
LGTM 👍

PR #1156, opened 2022-11-17
Initial dev docker setup

jabrah:
* Remove all of the previous docker development setup
* Add new docker image that slots into pass-docker's demo setup
  * Based on https://github.com/danlynn/ember-cli with some customizations to fit our needs (mostly being able to define the port at which the Ember app is served)

Could use feedback on its usefulness on Mac.

## Testing
I pushed the dev image to GHCR, so it should be pulled down and be useful without the full rebuild.

As long as you have pass-docker in place as described below, you should be able to run:

``` sh
yarn run dev
```

This will spin up the demo environment with a new dev-friendly image for the UI which will run live code from pass-ui. Changes will automatically be re-built and available after manually refreshing the page in your browser.

## Issues

At the moment, this requires a specific directory structure. `pass-docker` and `pass-ui` need to be next to each other:

```
|- pass-docker/
\- pass-ui/
```

This is needed because of the way two aspects are handled in this iteration. The convenience script brings the demo's compose file in as the base file, which defines the base path when defining the location of the bind mounts. The issue is that we want to use the demo environment as a base and bring the new compose file to override some `pass-ui` settings so that we can insert our new image in place of the original pass-ui image in order to get the networking to function.

I'm not sure if there's a simple way to use dynamic values (like env vars) in the compose file to define bind mounts


markpatton (commented):
I had a problem with yarn:

```
 yarn run build:dev 
yarn run v1.22.17
error Command "build:dev" not found.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

Also docker-compose reference .demo_env but no such file


Instead I ran
```
touch .demo_env
./build.sh .demo_env.
````

Also build.sh should have $ENV_FILE quoted.

The build failed at docker
```
Step 11/12 : COPY ./dist/ /usr/share/nginx/html/app/
COPY failed: file not found in build context or excluded by .dockerignore: stat dist/: file does not exist
```






Should remove .docker?

jabrah:
The `build.sh` script is unrelated to this development setup, but the error you include here is likely because of the version of Node you're using. I also see this error when using Node v16+, so we have to stick to Node v14 until we can update ember-cli to a version that includes this fix

Thanks for pointing out that `yarn run build:dev` doesn't work, turns out I had just forgotten to push the one commit that adds that command

jabrah:
Do we think it would be better to have the UI dev environment managed here in `pass-ui`, or would it be better to move it into `pass-docker`? I don't think this would require many changes, mostly just the dev compose file would just point to where you have `pass-ui` locally on your file system in order to mount it to the dev container

jaredgalanis:
> Do we think it would be better to have the UI dev environment managed here in `pass-ui`, or would it be better to move it into `pass-docker`? I don't think this would require many changes, mostly just the dev compose file would just point to where you have `pass-ui` locally on your file system in order to mount it to the dev container

Personally, it seems reasonable to have this dev env live here in pass-ui. pass-docker, unless we change something, is less of a dev env and more of a portable running instance of the stack.

jaredgalanis (approved):
:shipit:

jabrah:
Note that this setup must change slightly when/if https://github.com/eclipse-pass/pass-docker/pull/289 is merged

jaredgalanis:
Discussed and we are opting to develop pass-ui using `MIRAGE='true' ember s` instead

PR #1170, opened 2023-03-24
Fix styling post ember 4 upgrade

jaredgalanis:
There were various styling issues when running the application locally that needed fixing and weren't surfaced in the tests due to a different `index.html`.

Screenshot 2023-03-24 at 4 12 17 PM


jabrah (approved):

jabrah (approved):
We walked through these changes together. Looks good

PR #1172, opened 2023-03-27
fix: do not clobber alert styles in coreUI with flash message styles

jaredgalanis:


jabrah (approved):
Ah, at least a workaround is fairly simple

Would be nice to have some sort of visual regression testing at some point in the future to catch these sorts of issues

PR #1179, opened 2023-04-18
Snapshot versions should build dev images

jabrah:
"Build" action can be told to build dev or prod artifacts. Dev build is default behavior, Release workflow updated to make "prod" image for full release, but dev version for any -snapshots


jaredgalanis (approved):
:shipit:

PR #1180, opened 2023-04-20
Remove schema service mocking, update service URL path

jabrah:
Seems to work to make requests, but the service in pass-core is not ready yet


jaredgalanis (approved):
Approving conditionally while tests are being fixed

PR #1191, opened 2023-06-22
fix: enable prod builds

jaredgalanis:
- we need to not fingerprint assets for now or it will break the theme'ing (we'll fix that later)
- we also need to not strip the data test selectors because test cafe relies on them


jabrah (approved):
Won't see any effects except for production builds, which are only built and created on full releases. Local testing w/prod build looked fine

PR #1192, opened 2023-07-05
Allow workflows to call this release workflow

jabrah:
Needed if we consolidate releases into a single workflow in `eclipse-pass/main`, would be called like: https://github.com/eclipse-pass/main/blob/80a99993637d93708ff354c28844ea78ced54b82/.github/workflows/release.yml#L33-L37


cshannon2 (approved):
Looks good to me.

PR #1194, opened 2023-07-18
feat: add content security policies

jaredgalanis:


jabrah (approved):
Would be nice to remove the `unsafe-inline` and `unsafe-eval` down the line, but this is a good start

PR #1223, opened 2023-10-09
Externalize instructions Url to config.json

rpoet-jh:
This PR externalizes the instruction URL in the notice banner to the config.json so it can vary by env.

https://github.com/eclipse-pass/main/issues/690


jaredgalanis (approved):
Looks good! 👍

PR #1224, opened 2023-10-12
fix: add better policy step tests

jaredgalanis:


tests/unit/models/journal-test.js
```diff
@@ -0,0 +1,22 @@
...
+module('Unit | Model | journal', (hooks) => {
+  setupTest(hooks);
+
+  test('it exists', function (assert) {
+    const store = this.owner.lookup('service:store');
+    const journal = store.createRecord('journal');
+    assert.ok(journal);
+  });
+
+  test('it computes pmcParticipation correctly', function (assert) {
```
jaredgalanis, line 13: this might be the only test that was missing entirely that addresses that regresssion

mirage/config.js
```diff
@@ -285,7 +285,10 @@ export default function (config) {
     issns: ['Print:0003-2654', 'Online:1364-5528'],
     journalName: 'The Analyst',
     nlmta: 'Analyst',
+    pmcParticipation: 'B',
```
jaredgalanis, line 288: We need to add this explicitly because otherwise, this default https://github.com/eclipse-pass/pass-ui/pull/1224/files#diff-cd49b43c6fe4ed8bd896068427733aa3dab8fb10d917ab6bf67a0c325b0a6811R7 in the factory overrrides it.

tests/integration/components/policy-card-test.js
```diff
@@ -1,32 +1,39 @@
...
+    this.store = this.owner.lookup('service:store');
+  });
 
   test('it renders when given expected data', async function (assert) {
-    const policy = EmberObject.create({
-      repositories: A(),
-      description: 'This is a moo-scription',
-      title: 'Moo title',
-      id: 1,
+    const submissionAttrs = this.server.create('submission').attrs;
```
jaredgalanis, line 16: Pretty much all of the changes in this file involve moving to mirage and factories for the mock data and then also using the new data test selectors to be more declarative about what is being tested.

jabrah (approved):

PR #1225, opened 2023-10-12
Fix pagination of proxy user search component

markpatton:
This pr corrects the pagination with elide and makes sure we grab the total pages from the meta block of the result.
In addition the user query is normalized so that special characters don't make the query invalid.


jaredgalanis (approved):
Looks good to me. The only thing I'd suggest is removing the two newly generated but empty files. I'm guessing that's some kind of auto-generated behavior based on your editor or something.

As you have mentioned, this is currently a bit difficult to test manually given the lack of users in the db. I can follow up with a PR for an automated test if you want.

markpatton:
Oh jeez. I somehow committed some empty files. Will fix.

PR #1226, opened 2023-10-13
Call user service with a userToken parameter

markpatton:
Makes sure the user service gets called once if the userToken parameter is provided before everything else. The service may modify submission objects in the backend.

This also removes the weird route munging which was needed because of proxy url encoding issues and is no longer needed.


jaredgalanis (approved):
🚀

PR #1231, opened 2023-10-19
fix: deposit requirement modal copy

jaredgalanis:
Screenshot 2023-10-19 at 10 36 36 AM

Closes: https://github.com/eclipse-pass/main/issues/402

### DIFF
--- a/app/components/workflow-review/index.css
+++ b/app/components/workflow-review/index.css
@@ -35,5 +35,9 @@ table td {
 
 .comment-text-area {
   resize: none;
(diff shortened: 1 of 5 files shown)


jabrah (commented):
Seems fine to me to clarify the actions here, but I'm not sure if it fully resolves the UX point raised in the ticket, since IMO it's still as easy to breeze past the checkbox without reading the label

jabrah (approved):
Looks good once tests are fixed

PR #1232, opened 2023-10-19
Fix bug displaying error message when doi lookup fails

markpatton:
If you paste in a valid doi and then just mess with it by adding  or deleting a character, you get an error popup. This is just caused by a bug handling the error message.


jaredgalanis (approved):

markpatton:
Done.

PR #1234, opened 2023-10-19
fix: publication title not updating after amending it

jaredgalanis:
Closes: https://github.com/eclipse-pass/main/issues/763

### DIFF
--- a/app/controllers/submissions/new/basics.js
+++ b/app/controllers/submissions/new/basics.js
@@ -65,7 +65,7 @@ export default class SubmissionsNewBasics extends Controller {
 
   @action
   async loadTab(gotoRoute) {
-    if (!this.doiInfo.title) set(this, 'doiInfo.title', get(this, 'publication.title'));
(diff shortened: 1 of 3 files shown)


app/controllers/submissions/new/basics.js
```diff
@@ -65,7 +65,7 @@ export default class SubmissionsNewBasics extends Controller {
 
   @action
   async loadTab(gotoRoute) {
-    if (!this.doiInfo.title) set(this, 'doiInfo.title', get(this, 'publication.title'));
```
jaredgalanis, line 68: This is the actual change that fixed things for the bug.

jaredgalanis: Note, stashing this information on a piece of Doi Info is something maybe we should consider changing earlier. It took me a while to even hunt this down in part because of the naming.

Also note, you can proceed initially with a DOI and then backtrack and remove it and use a manually inputted title and journal and this still all works.

jabrah (approved):

PR #1236, opened 2023-10-23
Add helper text for journal selection.

markpatton:


app/components/workflow-basics/index.hbs
```diff
@@ -150,6 +150,11 @@
 </div>
 
 <div class="form-group">
+  <p class="lead text-muted">
+    If you are an NIH-user, a journal selection is required to deposit into PMC. It is optional for all other users. If
+    you do not see your journal listed, you can skip the journal selection and still use PASS to deposit into
+    JScholarship.
```
jaredgalanis, line 156: Seems good to me for now. I think down the road this will make less sense for non-JHU implementations.

jaredgalanis (approved):

PR #1241, opened 2023-10-25
fix: remove compliance column in grants index table

jaredgalanis:
per discussions on slack we're removing this column until we can better determine what policy compliance means


rpoet-jh (approved):

markpatton (approved):
I ran it locally and the grants page worked fine without the column.

PR #1243, opened 2023-11-09
Adding server-side pagination support

jabrah:
(Will) Close https://github.com/eclipse-pass/main/issues/547

## Changes

* Grants list pagination largely working, global filter not hooked up

## Description

Adding support for server-side pagination, where previously we grabbed all records and did client side paging. For now, I've added some URL query params: 

* `page` [integer] : current page to show
* `pageSize` [integer] : number of items per page
* `filter` [string] : global search string to narrow items shown on the page
  * Currently only over `grant.projectName`

These query params will be automatically added to the URL and updated as the user uses the pagination controls, but the user can also manually update these parameters in their address bar, if desired

## TODO

* [x] Connect global search/filter on grants list
  * Should investigate using an index for this in a separate PR
* [x] Add pagination on submissions page
* [x] Add pagination to the grant details page
* ~[ ] Add pagination to the grant selector in the submission workflow~
  * Look into in a separate ticket, since the UI is different here
* ~[ ] Debounce requests as needed for the above implementations~
  * ~Need to see about separating debounce of pagination params vs filter (out-of-the-box they are handled with the same mechanism)~
  * ember-models-table should debounce requests automatically. E.g. global filter is handled here https://github.com/onechiporenko/ember-models-table/blob/f8d00579aa6684bcf0e874c384e03b508ec79ade/addon/components/models-table-server-paginated.ts#L321-L324 (link is to older v4.8.0, which is what is used in PASS currently). The request is handled by the `_loadDataOnce` which handles debouncing via the Ember runloop. Default is 500ms
* [ ] Try to refactor queries to reduce code duplication?
* [x] Need to handle global search strings better (e.g. spaces currently break the query)

## Questions

* How does the UI respond to invalid or out-of-bounds query params? E.g. setting `page=20` when there's only 1 page of results from the backend?
* What properties do we search over for the given pagination situations?
  * Related: We should probably setup indexes in the backend to support these filters/searches

### Manual
You can check any of the above sections. Using pagination controls will update URL query parameters on the page.
Since we don't have a huge amount of test data, you can manually set page size to something small, like `pageSize=2` to force the results to render across multiple pages

### DIFF
--- a/app/controllers/grants/detail.js
+++ b/app/controllers/grants/detail.js
@@ -1,64 +1,137 @@
 /* eslint-disable ember/no-classic-classes */
 import Controller from '@ember/controller';
+import { action } from '@ember/object';
+import { tracked } from '@glimmer/tracking';
 import { inject as service } from '@ember/service';
-import { computed } from '@ember/object';
-import { alias } from '@ember/object/computed';
+import { hash } from 'rsvp';
 
-export default Controller.extend({
-  currentUser: service('current-user'),
-  themeInstance: service('emt-themes/bootstrap4'),
+export default class GrantDetailsController extends Controller {
+  @service('emt-themes/bootstrap4') themeInstance;
+  @service currentUser;
+  @service store;
 
-  grant: alias('model.grant'),
+  get grant() {
+    return this.model.grant;
+  }
 
   // Columns displayed depend on the user role
-  columns: computed('currentUser', {
-    get() {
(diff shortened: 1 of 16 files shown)


jabrah:
We could introduce an RSQL library if we wanted to abstract the RSQL query generation better

jaredgalanis (approved):
🎉

PR #1248, opened 2024-01-17
Delete files when removed from Draft submissions

jabrah:
Resolves https://github.com/eclipse-pass/main/issues/514

- Update the `deleteFile` function in the Files step to call the file service to delete the file, then delete the File object in the DB
- Add console log in the same place to print when an error occurs
- Updates to tests
  - Check to make sure `file.destroyRecord` is called
  - General test code updates

### DIFF
--- a/app/adapters/file.js
+++ b/app/adapters/file.js
@@ -1,0 +1,12 @@
+import ApplicationAdapter from './application';
+
+export default class FileAdapter extends ApplicationAdapter {
+  deleteRecord(store, type, snapshot) {
+    console.warn('File deletion cannot be rolled back');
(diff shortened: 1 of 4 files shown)


app/adapters/file.js
```diff
@@ -0,0 +1,12 @@
+import ApplicationAdapter from './application';
+
+export default class FileAdapter extends ApplicationAdapter {
+  deleteRecord(store, type, snapshot) {
+    console.warn('File deletion cannot be rolled back');
```
jabrah, line 5: This probably doesn't need to be put in the console. Should I move to a code comment?

jaredgalanis: I think this is fine.

jaredgalanis (approved):

PR #1252, opened 2024-01-29
refactor: update readme

jaredgalanis:
Closes: https://github.com/eclipse-pass/main/issues/770


jabrah (approved):
Look fine to me. Removing a bunch of the unnecessary detail will probably make it less of a burden to keep up to date. Interested to see if someone not deep in pass-ui would need any more instruction on running and testing though

PR #1257, opened 2024-02-27
Force config loading at application root

jabrah:
Fixes https://github.com/eclipse-pass/main/issues/929

# Changes

- Move static config loading from `afterModel` hook to `beforeModel` hook
  - Force waiting for this to resolve before proceeding with any/all routes to avoid errors in those routes when trying to access the config
- Add basic regression test to make sure routes outside of root load directly
- Also add a `console.error` to dump error messages into JS console in browser on route/model hook errors
  - The error case presented by this bug should at least be dumped into the browser console with this

# Testing

- You'd need to create a local build of this branch of `pass-ui`
  - The included `build.sh` can help: `> ./build.sh <path_to_pass-docker>/.env`
  - This will produce a new Docker image: `ghcr.io/eclipse-pass/pass-ui:latest`
- In `eclipse-pass.local.yml` in `pass-docker`, update the image ref to use the `latest` tag that you just built
- Bring services in pass-docker up as normal
- Navigate to `pass.local` and go through the regular login process
- Click into the Submissions page (`/app/submissions`)
- Refresh the page
- The page should load as expected
  - Previously, this refresh would hang permanently with a loading spinner and never resolve

### DIFF
--- a/app/routes/application.js
+++ b/app/routes/application.js
@@ -23,8 +23,12 @@ export default class ApplicationRoute extends CheckSessionRoute {
    * If there is a userToken query parameter call the user service with that parameter
    * to ensure objects are updated in the backend before any queries are done.
    */
-  beforeModel(transition) {
+  async beforeModel(transition) {
     let userToken = transition.to.queryParams.userToken;
+
+    if (!this.staticConfig.config) {
+      await this.staticConfig.setupStaticConfig();
+    }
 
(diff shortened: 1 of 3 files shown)


rpoet-jh (approved):
Good job, John.  I tested locally following your instructions, and it refreshed successfully.

PR #1262, opened 2024-04-25
refactor: don't call authenticate on every visit to check session

jaredgalanis:
- The UI was sending a request to whoami on every route visit
- This avoids that by only authenticating if a session doesn't already exist


markpatton (approved):
I tested locally and it works.

PR #1264, opened 2024-05-06
fix: session data shape bug

jaredgalanis:
Closes: #https://github.com/eclipse-pass/main/issues/974

### DIFF
--- a/app/authenticators/http-only.js
+++ b/app/authenticators/http-only.js
@@ -1,4 +1,3 @@
-/* eslint-disable ember/no-computed-properties-in-native-classes */
 import Base from 'ember-simple-auth/authenticators/base';
 import RSVP from 'rsvp';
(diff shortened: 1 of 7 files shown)


markpatton (changes requested):
The code looks good as well as the test.

But there does seem to be one regression. I'm noticing that the whoami service is getting called on every click.

jaredgalanis:
> The code looks good as well as the test.
> 
> But there does seem to be one regression. I'm noticing that the whoami service is getting called on every click.

@markpatton, it was the new `setup` method that's required after bumping ember-simple-auth that was calling restore every time. 

I was not transitioning in my testing because I was running without the backend. The latest change just moves where setup is called until after we have verified that we are not authenticated.

markpatton (approved):
Looks good. I don't see multiple whoami calls anymore.

PR #1265, opened 2024-05-16
Ignore pushes to main with tags in workflows

rpoet-jh:
Needed for complete workflow PR: https://github.com/eclipse-pass/main/pull/980


markpatton (commented):
Looks good.

PR #1266, opened 2024-05-17
Update gh workflows for complete release workflow

rpoet-jh:
Needed for complete workflow fixes PR: https://github.com/eclipse-pass/main/pull/983

Updated ci workflow to trigger on same events as pass-core/pass-support.

Changed update docker images workflow to trigger on PR merge.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,13 +1,5 @@
-name: CI
-
-on:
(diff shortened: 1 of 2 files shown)


markpatton (approved):
Only thought is will that if condition pass on a manually triggered job?
Should it check for ```
 github.event_name == 'workflow_dispatch' 
```

PR #1268, opened 2024-05-25
refactor: remove non strict and non explicit relationships deprecation

jaredgalanis:
This one involves the following deprecations:

https://deprecations.emberjs.com/ember-data/v4.x#toc_ember-data-deprecate-non-strict-relationships

This one we already got for free by upgrading ember simple auth in the process of the auth refactoring:

ember-simple-auth.initializer.setup-session-restoration


markpatton (approved):
I noticed these warnings in the console:

```
warn.js:59 WARNING: You pushed a record of type 'funder' with a relationship 'policy' configured as 'async: false'. You've included a link but no primary data, this may be an error in your payload. EmberData will treat this relationship as known-to-be-empty.

click @ index.js:1072
warn.js:59 WARNING: You pushed a record of type 'publication' with a relationship 'journal' configured as 'async: false'. You've included a link but no primary data, this may be an error in your payload. EmberData will treat this relationship as known-to-be-empty.
```

When I look at the funder.js and publication.js models I see they have the `async: false, inverse: null` which were removed from the others.

PR #1279, opened 2024-07-01
fix: allow proxy submission to submit without manuscript

jaredgalanis:


markpatton (approved):
I tried submissions with and without a journal and did not notice any problems.

PR #1281, opened 2024-07-05
fix: pass csrf token on doi/manuscript fetch and clean up withCredentials

jaredgalanis:


app/services/oa-manuscript-service.js
```diff
@@ -28,7 +28,13 @@ export default class OAManuscriptService extends Service {
 
     const url = `${this.lookUpPath}?doi=${doi}`;
 
-    return fetch(url, { method: 'GET' })
+    return fetch(url, {
```
jaredgalanis, line 31: This one was missing the csrf token and was being spiked by the DOI service.

markpatton (approved):
Looks good.  Ok. So the consequence of this is that the manuscript file fetch will just not return files until this fix is present. I think I had forgotten that functionality existed.

I tried it out and did get a 200 back from the service. But strangely the ui did not say a pdf was available despite a url I tried in the returned json resolving to a pdf. This was the json:

```
{
  "manuscripts": [
    {
      "url": "https://bmchealthservres.biomedcentral.com/track/pdf/10.1186/s12913-020-05409-w",
      "repositoryLabel": null,
      "type": "application/pdf",
      "source": "Unpaywall",
      "name": "s12913-020-05409-w",
      "isBest": true
    },
    {
      "url": null,
      "repositoryLabel": null,
      "type": "application/pdf",
      "source": "Unpaywall",
      "name": null,
      "isBest": false
    },
    {
      "url": "https://www.researchsquare.com/article/rs-19757/v1.pdf",
      "repositoryLabel": null,
      "type": "application/pdf",
      "source": "Unpaywall",
      "name": "v1.pdf",
      "isBest": false
    },
    {
      "url": "https://www.researchsquare.com/article/rs-19757/v2.pdf?c=1631848933000",
      "repositoryLabel": null,
      "type": "application/pdf",
      "source": "Unpaywall",
      "name": "v2.pdf?c=1631848933000",
      "isBest": false
    }
  ]
}
```

PR #1285, opened 2024-08-01
fix: css issues

jaredgalanis:
- there was an issue with the loading of coreUI.css from the index.html file in production builds
- this moves coreUI to be a bundled css file
- this also removes an invalid import of branding.css that was unrelated, but throwing a browser console error
- this also fixes broken loading of logos and other static config assets

addresses: [#1027 Bootstrap 5 css not loading in pass-ui production build](https://github.com/eclipse-pass/main/issues/1027)


markpatton (approved):
The branding.css warning is gone from the log.

The header is displayed. Should it have a logo? If so, the logo is missing.

I tested doing a submission and looked through all the tables. The formatting seems fine.

PR #1292, opened 2024-12-02
Create SBOM

rpoet-jh:
This PR has the following changes:

- Added cdxgen to create an SBOM and defined `create_sbom` in package.json scripts.
- Added command to copy SBOM file into Docker image.  I verified the generated SBOM is valid.
- Created `dependencies` in package.json and moved "runtime" dependencies  to this section.  _Please review these to ensure I have not missed any from devDependencies or have a dependency that is not a runtime dependency._
- Updated some dependencies linked in index.html.
- Changed the github action workflows to use the `eclipse-pass/main/.github/actions/node-build@main` action instead of the `build-pass-ui` action.  `build-pass-ui` has been removed.
- Removed the `latest` tag from docker images created by github action workflows. This is to be more consistent with other repo action snapshot/release workflows.

The way I have things configured now, an SBOM will be created and copied into the corresponding Docker image whenever a PR is merged and when PASS is released.

Note this PR should be merged after the main PR is merged: https://github.com/eclipse-pass/main/pull/1087


jaredgalanis (approved):
Looks good to me. I think there are some packages we could talk about moving, but I think in the interest of getting this started it's okay to include them as is for now.

For reference, here's what an app's `package.json` would look like if you did `ember new`, which might be instructive for us. https://github.com/ember-cli/ember-new-output/blob/v5.12.0/package.json

PR #1293, opened 2025-01-09
Set USER in docker

rpoet-jh:
This PR updates the Dockerfile so the pass-ui docker container runs with the non-privileged nginx user.

Note that the base image was changed to nginxinc/nginx-unprivileged:1.27.3-alpine.  This is what I found was the suggested way for running the nginx docker container with a non-privileged user. (https://hub.docker.com/_/nginx Running nginx as a non-root user).  Also note the upgrade in nginx version to 1.27.3.

I have tested this locally in pass-docker and acceptance tests pass.  I also tested the config where custom branding is mounted.  L&F changed and tests passed.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM nginx:1.22.0-alpine
+FROM nginxinc/nginx-unprivileged:1.27.3-alpine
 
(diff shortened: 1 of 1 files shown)


markpatton (approved):
Looks good.

PR #1294, opened 2025-01-13
Enable CSRF for logout

rpoet-jh:
This PR changes logout to POST to `/logout` when logout is clicked.

This PR goes with this `pass-core` PR, both need to merged at the same time.


markpatton (approved):
I tested locally in pass-docker and it worked well.

PR #1297, opened 2025-01-28
Fix publication date parsing of doi metadata 

markpatton:
The publication date formatting stored in the metadata must match what is expected by the data picker.
The error messages about missing metadata are now better formatted.

I was worried about  the publication date not being editable if it is not in the doi metadata since it is required. But despite it not seeming to be required, it is on every DOI I look at. And making a change so that missing values are editable also might be problematic.


jaredgalanis (approved):
LGTM 👍

PR #1301, opened 2025-03-12
First draft of replacing alpaca and removing old dependencies

markpatton:


app/controllers/submissions/detail.js
```diff
@@ -252,7 +252,8 @@ export default class SubmissionsDetail extends Controller {
     if (allLinksVisited) {
       this.hasVisitedWeblink = true;
     }
-    $('#externalSubmission').modal('hide');
+
+    document.querySelector('#externalSubmission').hide();
```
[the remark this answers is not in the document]
markpatton: @jaredgalanis  In this case is using `swal.close()` instead correct? I tested with the original code and with `swal.close()` and the behavior seems the same, everything works, in both cases.

rpoet-jh:
@markpatton Did some testing, here are my notes:

**Bugs:**

- Edit of submission: details form doesn't show publication date or author
- Form not updating attribute if grant/repos are changed in submission workflow.  
Steps:  
Login as nihuser  
create submission  
journal: Diabetes forecast  
grant: select QQDV123P7, next through to Details  
NTMLA attribute is displayed  
go back to grants page and remove QQDV123P7 and add TEST_E2E_AWD_NUM and click next through to Details. NTMLA attribute still in form.  If you select TEST_E2E_AWD_NUM on new submission, NTMLA attribute is not in form.

**Styling:**

- paginator layout is off
- buttons on Dashboard underline when hover
- table rows are striped

_workflow_
- buttons on Basics and Review steps too close to component above them (margin/padding off)
- Details
	- Font needs to be changed to match
	- Form should be left justified
	- Width of text fields should fill 100%
	- Should NLMTA be with Journal fields?
	- Make Complete button the Next button on the workflow button footer 

_Submissions_
- Buttons in the Actions columns need styling on table

_Submission Details_
- With Approval Requested submission, as the approver, click on submission link to open the details page.  The layout of the Repositories (2 repos) is vertical instead of horizontal.  I think the width of whole attribute panel needs to fill parent.
- Buttons on bottom too close to details panel

app/controllers/submissions/detail.js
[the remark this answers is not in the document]
markpatton: I think the right solution is using swal.close.

app/components/workflow-files/index.js
```diff
@@ -12,6 +13,13 @@ export default class WorkflowFiles extends Component {
   @service currentUser;
   @service flashMessages;
 
+  @tracked doi = null;
+
+  constructor() {
+    super(...arguments);
+    this.doi = this.args.doi;
```
jaredgalanis, line 20: I'm surprised that you needed the constructor here (as opposed to just setting the field above), but if it works I think it's fine to leave as is.

app/index.html
```diff
@@ -20,16 +20,6 @@
...
-    <script type="text/javascript" src="https://code.jquery.com/jquery-3.6.4.min.js"></script>
-    <script
-      src="https://cdn.jsdelivr.net/npm/bootstrap@4.0.0/dist/js/bootstrap.min.js"
-      integrity="sha384-JZR6Spejh4U02d8jOt6vLEHfe/JQGiRRSQQxSfFWpi1MquVdAyjUar5+76PVCmYl"
-      crossorigin="anonymous"
-    ></script>
-    <script type="text/javascript" src="https://cdnjs.cloudflare.com/ajax/libs/handlebars.js/4.7.8/handlebars.js"></script>
-    <script type="text/javascript" src="https://code.cloudcms.com/alpaca/1.5.24/bootstrap/alpaca.min.js"></script>
-
     {{content-for "body-footer"}}
```
jaredgalanis, line 23: 💯

app/components/metadata-form/index.js
```diff
@@ -1,47 +1,61 @@
...
-            componentContext.args.nextForm(stripEmptyArrays(this.getValue()));
-          },
-        },
+    const customCss = {
+      body: 'pt-4',
+      page: {
+        // Remove default survey.js page class
+        root: '',
+      },
+      navigation: {
```
jaredgalanis, line 20: I think we can get rid of this in favor of:

`this.survey.showCompleteButton = false;`

Below.

jaredgalanis (approved):
Looks good to me! Only a few comments that I don't think we need to address now.

PR #1303, opened 2025-03-24
Refactor submission files state management

rpoet-jh:
This PR makes the following changes to files state management:

- For the submission workflow files, file state is now held in the workflow service in an attribute named `files`.  The `files` state is set in the `submissions/new` route `model` function.  All other various file state attributes for submission workflow have been removed.
- Clean up `submission/detail` file state management.  The `model.files` is still being used for the detail view, just cleaned up loading and access functions in controller.

Notes:

- Re: setting the workflow services `files` attribute in the `submissions/new` route model function, I did testing to ensure when editing and creating new submission that route model function was properly triggered so that the `workflow.files` attribute is set.  I saw the `LinkTo @query` annotation is used to trigger these transitions.  I read in the ember doc that using `LinkTo @query` is not full-transition by default (meaning route model is not called), and I don't see `refreshModel` opt-in config.  However, in my testing, I see the route `model` function being called every time.  Maybe there is something else in the pass-ui config causing model to be called.  Please double check this.
- This PR fixes the remove file and other file issue noted in the the linked issue.  I did quite a bit of manual testing, but additional testing would be good to confirm these changes don't cause regression.


rpoet-jh:
I found an issue with additional testing in the proxy scenario on the details page.  I'll commit a fix soon, but it won't affect the other components, it will be isolated to the details component.

rpoet-jh:
Ok, all cleanup commits are done.  This is ready for review assuming CI finishes successfully.

rpoet-jh:
> Ok, all cleanup commits are done. This is ready for review assuming CI finishes successfully.

Some strange error on the Tests: `TypeError: $(...).modal is not a function`.  Not sure why this just started breaking, will look into tomorrow.

markpatton (approved):
I did some manual testing of adding removing files from the submission in various states and everything worked as expected.

rpoet-jh:
Thanks @markpatton the review and testing.  Before I merge, given the extent of the changes, it would probably be good to also get a review and testing from @jaredgalanis.

app/controllers/submissions/detail.js
```diff
@@ -28,6 +28,11 @@ export default class SubmissionsDetail extends Controller {
   @tracked externalRepoMap = {};
   @tracked _hasVisitedWeblink = null;
 
+  @computed('model.files')
```
jaredgalanis, line 31: Probably ideal not to add another `@computed` or Ember `get`, but I don't see this as a blocker in any way, more just an FYI.

I think the "new" cannonical way here would be to set up a tracked property like
`@tracked submissionFiles = this.model.files.slice();`

jaredgalanis (approved):
LGTM 👍 

one small observation. def non-blocking.

app/controllers/submissions/detail.js
rpoet-jh: Got it.  I'll make the change.

PR #1304, opened 2025-03-28
Fix page reload blank render issues

rpoet-jh:
This PR fixes a few issues with the UI rendering blank:

- Reload of page with query params such as filter on Submission page and refresh.
- In browser, enter valid path but invalid ID such as https://stage.pass.jhu.edu/app/submissions/999999.  Now shows 404 page.
- In browser, enter path like `/submissions/new/basicmmm` caused infinite page reload error.  Now shows 404 page.
- Other errors other than 401 and 403 on page load. Now shows dialog with error message.

### DIFF
--- a/app/controllers/grants/index.js
+++ b/app/controllers/grants/index.js
@@ -15,7 +15,7 @@ export default class GrantsIndexController extends Controller {
   constructor() {
     super(...arguments);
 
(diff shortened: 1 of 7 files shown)


markpatton (approved):
I did a bunch of manual testing and it looks good!

PR #1305, opened 2025-04-09
Fix submit confirmation bug

rpoet-jh:
This PR fixes the bug where if a submission is only being deposited into PMC, the submit confirmation dialog will now work as expected.  It also has some refactoring to try to make the conditional logic try a bit easier to understand.

Good to test confirmation dialog on Workflow Review page on Submit button, and the Submit button on the Submission Details page in the proxy workflow request approval use case.


markpatton (approved):
Looks good. I tried submitting to various combinations of repositories and everything worked. Looking at some of the code munging the agreement text, I wonder the data model around agreements should be cleaned up at some point.

PR #1306, opened 2025-05-01
Fix infinite error load issue on proxy submission details refresh

rpoet-jh:
This fixes an issue found during testing.  Here is when the issue happened:

- Running pass-docker locally
- Login as nih-user
- Start new submission
- Submit for proxy for staff1 user
- Go all the way through and Request approval
- Go to Submissions page and click the submission title link to view the details page
- Refresh the browser, this is where there would be an infinite error issue.

The issue was the `submitter` was not included when getting the submission.


markpatton (approved):
Tested locally. Looks good.

PR #1307, opened 2025-05-13
Configure SurveyJS not to load default Google fonts

markpatton:
This means that the fonts configured by the branding will be used by SurveyJS and that CSP rules for google fonts are not needed.


markpatton:
Huh. Just noticed that somehow this small change causes a test failure. Will investigate.

rpoet-jh (approved):
Looks good.  The status failure is the flaky accpt test, so approving.

PR #1309, opened 2025-05-30
Validate ORCID iDs

markpatton:
As an ORCID iD is entered in the metadata, its structure is validated. Both the long URL and short forms are supported. In addition a link formatted according to the guidelines, https://info.orcid.org/documentation/integration-guide/orcid-id-display-guidelines/, is added.

Note that having a user do full ORCID authentication to validate an ORCID iD does not make sense in the general case. A submitter or preparer would only be able to add their own ORCID iD.

TODO:
  * Normalize the ORCID iDs to the long form.
  * Decide if we should get an API key so we can actually lookup the person. What would we do with this info? Correct the name which was entered?

### DIFF
--- a/app/instance-initializers/surveyjs.js
+++ b/app/instance-initializers/surveyjs.js
@@ -1,0 +1,50 @@
+// Setup custom SurveyJS validation functions
+import { FunctionFactory } from 'survey-core';
+
+// Generates check digit as per ISO 7064 11,2.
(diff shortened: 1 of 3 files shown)


jaredgalanis (approved):
Tested manually LGTM 👍

markpatton:
@jaredgalanis Any thoughts on the link appearing after you type in a valid orcid? Does that seem reasonable?

PR #1310, opened 2025-07-07
Fix workflow grants pagination

rpoet-jh:
Now if the the user has more than 10 grants, the workflow grants page will show pagination controls allowing the user see all their grants.

There is also this pass-docker that adds grants for `nih-user` so they have more than 10 grants for testing purposes: https://github.com/eclipse-pass/pass-docker/pull/400


markpatton (approved):
Looks good. I can browse multiple pages of grants.

PR #1312, opened 2025-07-07
Fix Invalid DOI error message

rpoet-jh:
If a DOI had whitespace at the beginning or end, the regex would fail causing the "Invalid DOI" error message for a second until it was later trimmed after lookup.  This PR trims the DOI value before regex testing.


markpatton (approved):
I tried locally and the fix works!

PR #1313, opened 2025-07-08
Fix formatting of DSpace item repository ids

markpatton:
Handles the case of item URLs (produced by REST API deposit) and handle URLs (produced by old SWORD deposit)


markpatton:
@rpoet-jh 
The copilot review made the same comment! Will add.

